PR #3, opened 2016-01-13
Don't keep creating new accounts if a user exists. Instead, create dummy accounts with reproducible strings so we can find it again later.

cjcodes:


tongxiang:
:+1:

PR #5, opened 2016-01-14
add cron job to RecurShifts, extract time and location constants to config.js

tongxiang:
#### What's this PR do?
Adds a cron job to RecurShifts, extract time and location constants to config.js. (Note that we've committed `time_interval` and `locationID` constants to the `config.js.default` config template file, since it's not insecure to publish these. 

Has not yet been tested. 

Since all cron jobs run first on start, start the app to test if shift recurrence, first shift notification, and time off request auto-approval works. 

#### Questions:
Is assigning `const CONSTANT_NAME = global.config.locationID.variable_name` at the top of files superfluous? Should we just use references to `global.config` everywhere, instead of defining constant names? 

@andrewjgremmo--did I make the required changes in `TimeOffRequests.js` for prod?

### DIFF
--- a/config.js.default
+++ b/config.js.default
@@ -6,15 +6,50 @@ config.wheniwork = {
     api_key: 'TODO',
     username: 'TODO',
     password: 'TODO',
-    default_password: 'TODO',
-    new_graduate_location_id: 'TODO',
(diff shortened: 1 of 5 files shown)


andrewjgremmo:
:+1:

PR #7, opened 2016-01-14
remove `const`, add test environment changes to config.js

tongxiang:
#### What's this PR do?
Removes defined constants from all files, uses `global.config.XXXX.YYYY` variables instead. Also adds control flow in `config.js` which assigns a test location ID.


cjcodes:
:+1:

PR #10, opened 2016-01-14
Add OpenShift cloning.

cjcodes:


jobs/scheduling/OpenShifts.js
```diff
@@ -0,0 +1,68 @@
...
+    'published'
+];    
+
+new CronJob(global.config.time_interval.open_shifts, function () {
+    openShifts();
+}, null, true);
+
+function openShifts() {
+    var current_hour = moment().minute(0).second(0);
+    var next_time = current_hour.clone().add(7, 'days');
```
tongxiang: should we make this a constant to throw into `config.js`?

tongxiang: actually--what are we using this variable for?

jobs/scheduling/OpenShifts.js
```diff
@@ -0,0 +1,68 @@
...
+
+    // Example time filter: 1:59:00 pm - 2:01:00 pm
+    var filter = {
+        // start time = the top of the current hour -1 minute
+        start: current_hour.clone().add(-1, 'minute').format(wiw_date_format),
+        // end time = the top of the current hour +1 minute
+        end: current_hour.clone().add(1, 'minute').format(wiw_date_format),
```
tongxiang: shouldn't this be `next_time.clone().add(1, 'minute').format(wiw_date_format)`?

jobs/scheduling/OpenShifts.js
```diff
@@ -0,0 +1,68 @@
...
+
+                for (var j in copy_fields) {
+                    new_shift[copy_fields[j]] = shift[copy_fields[j]];
+                }
+
+                new_shift.start_time = moment(shift.start_time, wiw_date_format);
+                new_shift.end_time = moment(shift.end_time, wiw_date_format);
+
+                new_shift.start_time = new_shift.start_time.add(7, 'day').format(wiw_date_format);
+                new_shift.end_time = new_shift.end_time.add(7, 'day').format(wiw_date_format);
```
tongxiang: make this a constant?

tongxiang:
The cron currently only checks and replicates shifts for the specific hour it's running on (and it only runs once an hour). 

Because of this, if the job fails once for an hour, that hour will never be revisited and the shift will never be replicated to week 2. 

Week 3 also won't have that shift, since the shift hasn't been replicated to week 2. 

Our other jobs query for a large enough interval of shifts that if one job fails, the next time that job runs, it'll pick up previous shifts for which the job has failed and complete them successfully. 

We should do this, as well as including a validate function which confirms that the shift has been successfully replicated to the next week.

PR #16, opened 2016-01-15
Merge openshifts

cjcodes:
### Background
When we create openshifts for a time period for which open shifts already exist, those open shifts aren't always automatically incremented by one. Instead, multiple open shifts will then exist in that time slot. 

This is functionally harmless, but potentially confusing for the user. 

### What does this PR do? 
This PR finds time slots for which there are multiple open shifts, and merges those shifts into one shift. (While incrementing the quantity on that one shift, so that it can represent multiple shifts.)

### DIFF
--- a/jobs/scheduling/MergeOpenShifts.js
+++ b/jobs/scheduling/MergeOpenShifts.js
@@ -1,0 +1,65 @@
+
+var CronJob = require('cron').CronJob;
+var WhenIWork = require('./base');
+var moment = require('moment');
+
+var wiw_date_format = 'ddd, DD MMM YYYY HH:mm:ss ZZ';
+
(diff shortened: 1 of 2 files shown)


jobs/scheduling/MergeOpenShifts.js
```diff
@@ -0,0 +1,65 @@
+
+var CronJob = require('cron').CronJob;
+var WhenIWork = require('./base');
+var moment = require('moment');
+
+var wiw_date_format = 'ddd, DD MMM YYYY HH:mm:ss ZZ';
+
+new CronJob('0 0 * * * *', function () {
```
tongxiang: We should move out this cron string to `config.js`.

PR #22, opened 2016-01-15
Autologin for shift cancellation

cjcodes:
#### What's this PR do?
Currently, when a user goes to autologin to WhenIWork, we detect if it's an invalid username (their account wasn't created by us) and we redirect them to the login page accordingly.

However, this logic is not applying to the cancel shifts route. So, this PR adds the relevant `failureCallback` method to `api2`'s authentication call.

Attempt to cancel all shifts as a user with a manually created WiW account.

### DIFF
--- a/www/routes/scheduling/index.js
+++ b/www/routes/scheduling/index.js
@@ -85,10 +85,12 @@ router.get('/cancel-shift', function(req, res) {
                         console.log(response);
                     })
 
-                    var api2 = new WhenIWork(global.config.wheniwork.api_key, email, global.config.wheniwork.default_password);
+                    var api2 = new WhenIWork(global.config.wheniwork.api_key, email, global.config.wheniwork.default_password, function (resp) {
+                        res.render('scheduling/cancelShift', { url: 'https://app.wheniwork.com/' });
+                    });
 
                     api2.post('users/autologin', function (data) {
-                        res.render('scheduling/cancelShift', { token: data.hash });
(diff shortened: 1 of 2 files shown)


www/routes/scheduling/index.js
```diff
@@ -85,10 +85,12 @@ router.get('/cancel-shift', function(req, res) {
                         console.log(response);
                     })
 
-                    var api2 = new WhenIWork(global.config.wheniwork.api_key, email, global.config.wheniwork.default_password);
+                    var api2 = new WhenIWork(global.config.wheniwork.api_key, email, global.config.wheniwork.default_password, function (resp) {
+                        res.render('scheduling/cancelShift', { url: 'https://app.wheniwork.com/' });
```
tongxiang: If authentication fails (the user has a custom password that we haven't assigned), hits this failure callback and displays a view for the user to login with normally.

www/routes/scheduling/index.js
```diff
@@ -85,10 +85,12 @@ router.get('/cancel-shift', function(req, res) {
...
+                    });
 
                     api2.post('users/autologin', function (data) {
-                        res.render('scheduling/cancelShift', { token: data.hash });
+                        res.render('scheduling/cancelShift', { url: 'https://app.wheniwork.com/myschedule?al=' + data.hash });
```
tongxiang: otherwise, send this response.

tongxiang:
:+1:

PR #23, opened 2016-01-15
Add a destionation parameter to the autologin, allowing third party tools (like CTL Online) to specify where we want to redirect users after logging in.

cjcodes:


tongxiang:
:+1:

PR #28, opened 2016-01-27
add individual deletion of shifts

tongxiang:
#### What's this PR do?
1. adds the individual deletion of shifts functionality to Elmer. 
2. Also removes the bulk deletion route of shifts. 

#### What's the user flow? 
1. When the user navigates to online.crisistextline.org/schedules, she'll see a new button to the page which says "drop individual shifts" (to be added.) 
2. Clicking this button will GET the `/scheduling` route on Elmer, displaying a page which lists all the user's shifts in a checkbox form. Only two checkboxes are able to be checked at one time, roughly reflecting WhenIWork's API limit. 
3. When the user checks a shift's box and submits the form, we POST to the `/delete-shifts` route. 
4. The user is redirected to a new view: `someShiftsCancelled.jade`, which then gives a link for the user to be redirected back to When I Work, so that they can schedule another shift. 

Note below that the app crashes upon a successful delete shift request, so we can't confirm that the final page is displayed. 

Currently has been tested by scheduling two shifts in the `test` location on WiW, then deleting them together or individually with Elmer after commenting out the checks for a valid email and token. 
#### Background:
**note 1**
**TODO**: Currently, our successful delete-shift batch requests consistently return with an 502 error encoded in HTML in the body of the response. But this breaks line 159 within `index.js` of the module `wheniwwork_unofficial`. We need to resolve this--a crappy solution that might be the best solution would be to add a check for HTML on line 159. 

Currently, this crashing breaks the redirect after shifts are cancelled. 

**note 2**
Since the shift recurrence cronjob currently runs every minute, it's possible that a user books shifts and attempts to delete them before the shift recurrence cronjob has successfully recurred that shift. 

If this happens, the shifts won't delete--much of the shift deletion logic relies on the `notes.parent_shift` param, which is only added during shift recurrence--and instead the user will be redirected to an error message telling her to wait 30 seconds and then try again.

### DIFF
--- a/www/public/javascripts/main.js
+++ b/www/public/javascripts/main.js
@@ -1,0 +1,11 @@
+$(document).ready(
+	function(){
+	    var max = 2;
+	    var checkboxes = $('input[type="checkbox"]');
+
+	    checkboxes.change(function(){
+	        var current = checkboxes.filter(':checked').length;
+	        checkboxes.filter(':not(:checked)').prop('disabled', current >= max);
+	    });
+	}
+)
--- a/www/public/javascripts/main.js
+++ b/www/public/javascripts/main.js
@@ -1,14 +1,13 @@
-var max = 2;
+var maxCheckedBoxes = 2; // Only two boxes can be checked at a given time. Restricting because of perceived WhenIWork API limits.
 var checkboxes = $('input[type="checkbox"]');
 
 checkboxes.change(function(){
-    var current = checkboxes.filter(':checked').length;
-    checkboxes.filter(':not(:checked)').prop('disabled', current >= max);
+    var currentBoxesChecked = checkboxes.filter(':checked').length;
+    checkboxes.filter(':not(:checked)').prop('disabled', currentBoxesChecked >= maxCheckedBoxes);
 });
 
 $("#delete-shifts").click(function(e){
     e.preventDefault();
-
     var deleteURL = "/scheduling/shifts?" + $("form[name=cancel-form]").serialize();
 
     $.ajax({
--- a/www/public/javascripts/main.js
+++ b/www/public/javascripts/main.js
@@ -1,11 +1,7 @@
-$(document).ready(
-	function(){
-	    var max = 2;
-	    var checkboxes = $('input[type="checkbox"]');
+var max = 2;
+var checkboxes = $('input[type="checkbox"]');
 
-	    checkboxes.change(function(){
-	        var current = checkboxes.filter(':checked').length;
-	        checkboxes.filter(':not(:checked)').prop('disabled', current >= max);
-	    });
-	}
-)
+checkboxes.change(function(){
+    var current = checkboxes.filter(':checked').length;
+    checkboxes.filter(':not(:checked)').prop('disabled', current >= max);
+});
--- a/www/public/javascripts/main.js
+++ b/www/public/javascripts/main.js
@@ -5,3 +5,17 @@ checkboxes.change(function(){
     var current = checkboxes.filter(':checked').length;
     checkboxes.filter(':not(:checked)').prop('disabled', current >= max);
 });
+
+$("#delete-shifts").click(function(e){
+    e.preventDefault();
+
+    var deleteURL = "/scheduling/shifts?" + $("form[name=cancel-form]").serialize();
+
+    $.ajax({
+        url: deleteURL,
+        type: "DELETE",
+        success: function(data) {
+            window.location = data.redirect;
+        }
+    });
+});
--- a/www/routes/scheduling/index.js
+++ b/www/routes/scheduling/index.js
@@ -9,6 +9,139 @@ var router = express.Router();
 var api = new WhenIWork(global.config.wheniwork.api_key, global.config.wheniwork.username, global.config.wheniwork.password);
 
 var wiw_date_format = 'ddd, DD MMM YYYY HH:mm:ss ZZ';
+var choose_shift_to_cancel_page_start_date_format = 'dddd h:mm a';
+var choose_shift_to_cancel_page_end_date_format = 'h:mm a z'
+
+router.get('/', function(req, res) {
+    var email = "tong@crisistextline.org"
+    // var email = req.email;
+    // if (!validate(req.query.email, req.query.token)) {
+    //     res.status(403).send('Access denied.');
+    // }
+
+    api.get('users', function (users) {
+        users = users.users;
+        for (var i in users) {
+            if (users[i].email == email) {
+                var userID = users[i].id;
+                var query = {
+                    user_id: userID,
+                    start: moment().add(-1, 'day').format('YYYY-MM-DD 00:00:00'),
+                    end: moment().add(8, 'days').format('YYYY-MM-DD HH:mm:ss'),
+                    location_id: global.config.locationID.regular_shifts
+                };
+
+                api.get('shifts', query, function(response) {
+                    if (!response.shifts || !response.shifts.length) {
+                        var error = "You don't seem to have booked any shifts to delete! If this message is sent in error, contact scheduling@crisistextline.org";
+                        var scheduleShiftsURL = 'https://app.wheniwork.com/login/?redirect=myschedule'
+                        res.render('scheduling/chooseShiftToCancel', { error: error , url: scheduleShiftsURL});
+                        return;
+                    }
+                    var shifts = response.shifts;
+                    
+                    // Remove duplicate shifts returned
+                    var i = shifts.length;
+                    while (i--) {
+                        shifts = shifts.filter(
+                            function(shift, index) {
+                                if (i === index) {
+                                    return true;
+                                }
+                                return !areShiftsDuplicate(shifts[i], shift);
+                            }
+                        )
+                    }
+
+                    // Formatting shift time display to be more user-readable
+                    shifts.forEach(function(shift) {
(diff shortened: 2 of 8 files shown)


tongxiang:
502: frontend servers (proxy) is timing out after 10 seconds, so returning 502 errors. They did this in order to fix their issues on Friday when the server went down. 

Looks like load balancer timeout isn't in sync with the backend.

www/routes/scheduling/index.js
```diff
@@ -9,6 +9,151 @@ var router = express.Router();
...
+router.get('/', function(req, res) {
+    var email = req.email;
+    if (!validate(req.query.email, req.query.token)) {
+        res.status(403).send('Access denied.');
+    }
+
+    api.get('users', function (users) {
+        users = users.users;
+        for (var i in users) {
+            if (users[i].email == email) {
```
cjcodes: WHat if it's a funky admin+ email?

www/routes/scheduling/index.js
```diff
@@ -9,6 +9,151 @@ var router = express.Router();
...
+                var userName = users[i].first_name + ' ' + users[i].last_name;
+                var userID = users[i].id;
+                var query = {
+                    user_id: userID,
+                    start: moment().add(-1, 'day').format('YYYY-MM-DD 00:00:00'),
```
cjcodes: `start: '-1 day', end: '+8 days'` see PHP `date()`

www/routes/scheduling/index.js
```diff
@@ -9,6 +9,151 @@ var router = express.Router();
...
+                    res.render('scheduling/chooseShiftToCancel', { shifts: shifts, userID: userID, email: email, token: req.query.token, userName: userName });
+                })
+                break;
+            }
+        }
+    })
+})
+
+function areShiftsDuplicate(shiftA, shiftB) {
+    return JSON.parse(shiftA.notes).parent_shift === JSON.parse(shiftB.notes).parent_shift;
```
cjcodes: If JSON.parse no worky?

www/routes/scheduling/index.js
```diff
@@ -9,6 +9,151 @@ var router = express.Router();
...
+    var parentShiftIDsOfShiftsToBeDeleted = [];
+    for (key in req.body) {
+        if (req.body[key] === 'on') {
+            parentShiftIDsOfShiftsToBeDeleted.push(parseInt(key));
+        }
+    }
+    
+    var query = {
+        user_id: req.body.userID,
+        start: moment().add(-1, 'day').format('YYYY-MM-DD 00:00:00'),
```
cjcodes: php date stuff

www/routes/scheduling/index.js
```diff
@@ -49,7 +195,7 @@ router.get('/cancel-shift', function(req, res) {
         for (var i in users) {
             if (users[i].email == email) {
                 var q = {
-                    user_id: users[i].id,
+                    user_id: userID,
```
cjcodes: userID not set anywhere.

www/views/layout.jade
```diff
@@ -3,5 +3,7 @@ html
   head
     title= title
     link(rel='stylesheet', href='/stylesheets/style.css')
+    script(type='text/javascript', src='https://code.jquery.com/jquery-2.2.0.min.js')
```
cjcodes: Move to bottom of body, and remove $(document).ready()

www/views/scheduling/chooseShiftToCancel.jade
```diff
@@ -0,0 +1,23 @@
+extends ../layout
+
+block content
+    - if (error) {
```
cjcodes: Should this be handled in an error.jade file somewhere? And we render an error page instead? I don't know Jade standards well enough to have a recommendation here.

tongxiang: Folks use Jade error pages for catching standard HTTP errors. I think this is fine.

tongxiang: http://stackoverflow.com/questions/10830183/handle-jade-errors-in-res-render

www/public/javascripts/main.js
```diff
@@ -0,0 +1,7 @@
+var max = 2;
```
indspenceable: call this `maxCheckedBoxes` or something? maybe a comment saying you can only check two boxes?

www/routes/scheduling/index.js
```diff
@@ -9,6 +9,198 @@ var router = express.Router();
 var api = new WhenIWork(global.config.wheniwork.api_key, global.config.wheniwork.username, global.config.wheniwork.password);
 
 var wiw_date_format = 'ddd, DD MMM YYYY HH:mm:ss ZZ';
+var choose_shift_to_cancel_page_start_date_format = 'dddd h:mm a';
```
indspenceable: oo dat variable name. But camelcase to be consistent w/ other names?

www/routes/scheduling/index.js
```diff
@@ -9,6 +9,198 @@ var router = express.Router();
...
+    if (!validate(req.query.email, req.query.token)) {
+        res.status(403).send('Access denied.');
+    }
+
+    var altEmail = email.replace(/\W+/g, '');
+    altEmail = 'admin+'+altEmail+'@crisistextline.org';
+
+    api.get('users', function (dataResponse) {
+        var users = dataResponse.users;
+        for (var i in users) {
```
indspenceable: be consistent - while both `for(v in list)` and `forEach(...)` are both valid, it's confusing to switch between them. forEach seems better IMO.

www/routes/scheduling/index.js
```diff
@@ -9,6 +9,198 @@ var router = express.Router();
...
+
+    var query = {
+        user_id: req.body.userID,
+        start: '-1 day',
+        end: '+8 days',
+        unpublished: true,
+        location_id: global.config.locationID.regular_shifts
+    };
+
+    api.get('shifts', query, function (shifts) {
```
indspenceable: this variable is actually `data`

www/routes/scheduling/index.js
```diff
@@ -9,6 +9,198 @@ var router = express.Router();
...
+        var parentShiftID
+          , shift
+          , batchPayload = []
+          , deletedShiftInformation = {}
+          ;
+
+        shifts.shifts.forEach(function(shift) {
+            parentShiftID = JSON.parse(shift.notes).parent_shift;
+
+            if (parentShiftIDsOfShiftsToBeDeleted.indexOf(parentShiftID) >= 0) {
```
indspenceable: why is this `>= 0` instead of `!= -1`

www/routes/scheduling/index.js
```diff
@@ -9,6 +9,198 @@ var router = express.Router();
...
+}
+// Route which allows individual deletion of shifts
+router.post('/shifts', function(req, res) {
+    if (!validate(req.body.email, req.body.token)) {
+        res.status(403).send('Access denied.');
+    }
+
+    var parentShiftIDsOfShiftsToBeDeleted = [];
+    for (key in req.body) {
+        if (req.body[key] === 'on') {
```
indspenceable: why do you do this in a `for` loop

www/routes/scheduling/index.js
```diff
@@ -9,6 +9,198 @@ var router = express.Router();
...
+    else if (firstMomentObject.minutes() !== secondMomentObject.minutes()) {
+        return firstMomentObject.minutes() > secondMomentObject.minutes() ? 1 : -1;
+    }
+    else if (firstMomentObject.seconds() !== secondMomentObject.seconds()) {
+        return firstMomentObject.seconds() > secondMomentObject.seconds() ? 1 : -1;
+    }
+    return 0;
+}
+// Route which allows individual deletion of shifts
+router.post('/shifts', function(req, res) {
```
indspenceable: this sounds like a `DELETE` not a `POST`

www/routes/scheduling/index.js
```diff
@@ -9,6 +9,198 @@ var router = express.Router();
...
+                if (!deletedShiftInformation[parentShiftID]) {
+                    var formattedStartTime = moment(shift.start_time, wiw_date_format).tz('America/New_York').format(choose_shift_to_cancel_page_start_date_format);
+                    var formattedEndTime = moment(shift.end_time, wiw_date_format).tz('America/New_York').format(choose_shift_to_cancel_page_end_date_format)
+                    deletedShiftInformation[parentShiftID] = { start_time: formattedStartTime, end_time: formattedEndTime };
+                }
+            }
+        })
+
+        api.post('batch', batchPayload, function(response) {
+            console.log('Shifts deleted response: \n', response);
```
indspenceable: This probably isn't necessary.

www/routes/scheduling/index.js
```diff
@@ -37,17 +229,20 @@ router.get('/login', function (req, res) {
     });
 });
 
+// Route to cancel all shifts
 router.get('/cancel-shift', function(req, res) {
```
indspenceable: `router.delete('/shifts/all')`?

www/routes/scheduling/index.js
tongxiang: Yup, thanks. And apparently `for i in array` iteration is an anti-pattern in JS. 

http://stackoverflow.com/questions/500504/why-is-using-for-in-with-array-iteration-such-a-bad-idea

cjcodes:
:+1:

PR #29, opened 2016-01-29
fix crash caused by timeoff requests not including shifts within their range

### LINKED ISSUE #26: confirm: will timeOffRequests crash the app if no pending time off requests are found?

tongxiang:
#### What's this PR do?
If a user creates a timeoff request, but she has not scheduled any shifts during that timeoff range, this no longer crashes the app. 

(previously would break at `response.shifts.forEach(function(shift) {`. 

Manually tested by creating a timeoff request not containing any shifts within its scope, and running. 

Closes #26.


cjcodes:
:+1:

PR #30, opened 2016-02-03
add proper redirect with query params for new shift scheduling

tongxiang:
### What does this PR do?

Adds proper redirects with query params for new shift scheduling. 

### Questions

Functionally, this PR provides the appropriate query params for redirects to new shift scheduling. This has only been tested with Andrew and I's credentials, though--admin accounts with separate passwords that WiW. And it hasn't been shown to work if we're logged out of WiW when we attempt the redirects. 

Will this work with a non-admin user? Need to test this, and to walk through the SSO token signin flow.


cjcodes:
:+1:

PR #32, opened 2016-02-05
Openshifts

tongxiang:
#### What's this PR do?
Previously, our `MergeOpenShifts` job exponentially increased the `instances` number on an openshifts. 

(If two openshifts had the same `instances` number, they'd both get updated with the new instances number and neither would be deleted. So if we begin with two open shifts, each with an instance number of 1, they'd both be updated to 2, then 4, then 8, then 16--each time the cron job runs once an hour.)

This fixes this problem with a flag. Also adds some camelCase, fixes whitespaces, and moves the cronjob string out into the global config.

### DIFF
--- a/config.js.default
+++ b/config.js.default
@@ -22,22 +22,22 @@ config.locationID = {
 }
 
 config.time_interval = {
-    // runs every 1 min. 
+    // runs every 1 min.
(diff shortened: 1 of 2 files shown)


cjcodes:
:+1: LGTM

PR #33, opened 2016-02-18
fix date format for finding shifts that fell within timeoff requests, was preventing them from being found

tongxiang:
### What does this PR do?
We were sending a moment date object instead of the proper date string when querying for shifts that should be changed given a time off request. 

This fixes that issue.


cjcodes:
:+1:

PR #35, opened 2016-02-26
add revised open shift recurrence

tongxiang:
#### What's this PR do?
Adds open shift recurrence based on a new model: 

1. Create a config file listing the number of shifts available for specific times (perhaps listed as a multiple of the number of `counselors_per_supervisor`)
2. revise recurrence job so that it runs every so often
3. master scheduling doc + WiW Shift Sums tab (link here: https://docs.google.com/spreadsheets/d/120A2OqHk5XG62-b4Xai0LeWiEBzpQbgmLpQRXvoP5XQ/edit#gid=685762792) 
4. What does function do? 
5. check config file
6. runs every two hours, pulls all shifts which start between 2 and 2:01 (start and end params are referring to the start and end times for the START TIME of the shift, not the whole shift itself.) 
7. finds all shifts for that time period for the regular shifts location (NOT including makeup shifts, which we want to save for Nan as a wildcard. So she can adjust the number of shifts all the time.) 
8. total - current_number_of_shifts = open shifts to be added to that same shift time next week. 
9. checks to see if open shifts have already been added to that time period. 
9. if those shifts haven't already been added, add those shifts.

It runs every two hours, duplicating previous shift, but also checking previous couple of shifts (for fail-safe-ing. In case the system goes down or misses an hour.) 

Check out config changes in `config.default.js` first. Then all the good stuff happens in `RecurOpenshifts.js`. 

Requires a few modifications to manually test. 
1. From `config.js.default`, add the objects for `numberOfSupervisorsPerShift` and `crisisCounselorsPerSupervisor` to the `config.js` file. 
2. Within `recurOpenShifts.js`, we need to specify which (even) hour we're running the script on. Comment out `var now = moment().minute(0).second(0);`, and add `var now = moment().hour(6).minute(0.second(0)`. 
3. Finally add a `recurOpenShifts()` function call. 

To run the entire Elmer app is to run a lot of the other jobs at the same time. To just run this job, we copy and paste the `config.js` file into `base.js` above the other code, along with `global.config = config` at the end of the config code. 

Finally, run `node /path/to/recurOpenShifts.js`. 

Tested lightly. Needs to be tested heavily before merge. 

Closes [Jira INT-3](https://admin.crisistextline.org/jira/browse/INT-3).

#### Questions:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -107,3 +107,4 @@ config.js
 !.elasticbeanstalk/*.global.yml
 
 *.csv
+*.swp
--- a/config.js.default
+++ b/config.js.default
@@ -19,7 +19,7 @@ config.locationID = {
     test2: 1007104,
     supervisors: 884473,
     crisis_counselors_demo: 1004215
-}
+};
 
 config.time_interval = {
     // runs every 1 min.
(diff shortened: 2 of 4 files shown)


cjcodes:
:+1:

PR #36, opened 2016-03-02
refactor open shift recurrence so that incorrect number of open shits are deleted

tongxiang:
#### What's this PR do?
Refactors open shift recurrence: 
1. We check to see if open shifts have already been recurred for the next week
2. Check if the number of open shifts matches up to the correct number of open shifts
3. If the numbers don't match, we delete the open shifts and add the correct number again

Tested by running job on test WhenIWork environment. 

Closes [JIRA INT-10](https://admin.crisistextline.org/jira/browse/INT-10). 
#### Questions:
1. Think through all delete shift / open shift creation cases. Think about if this works with edge cases.


tongxiang:
:+1:  <-- from Chris.

PR #37, opened 2016-03-03
Colorize recurred and merged shifts

cjcodes:
This PR should import a new library called ColorizeShift for use anywhere in the codebase. For now, it is required to explicitly call it. In the future, it may be possible to modify the wheniwork_nodejs library to support middleware, but this will do for now.

If a shift is defined in the config file, it will add a `color` property to the shift and return the new shift. Otherwise, it will just return the shift passed to it.

### DIFF
--- a/config.js.default
+++ b/config.js.default
@@ -80,5 +80,15 @@ config.numberOfSupervisorsPerShift = {
     'Sat' : { '12am': 2, '2am': 2, '4am': 1, '6am': 1, '8am': 2, '10am': 2, '12pm': 2, '2pm': 2, '4pm': 3, '6pm': 3, '8pm': 3, '10pm': 3}
 };
 
+var red = 'D61F27';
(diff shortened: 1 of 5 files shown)


cjcodes:
Library can be tested by running `node test/test.js` (until we build out a testing suite).

It should return:
```
{ start_time: 'Thu, 03 Mar 2016 04:00:00 -0500',
  color: 'D61F27' }
{ start_time: 'Thu, 03 Mar 2016 10:00:00 -0500' }
{ color: 'D61F27' }
{}
```

tongxiang:
Looks good. 

One more place to add `colorize`, though, could be in the `delete-shifts` route in `www/scheduling/index.js`. 

However, this open shift released when a user deletes a shift would be colorized when open shifts are merged. If you think this is enough, then :+1: .

PR #46, opened 2016-03-15
change time off job

tongxiang:
This PR does two things: 

1. Changes time off job to approve time off requests, even when they don't contain regular shifts. 
2. Adds functionality for taking time off which releases taken makeup shifts back into the makeup shift pool.


cjcodes:
:+1:

PR #48, opened 2016-03-16
change open shifts to modify open shift count if there are too many open shifts

tongxiang:
#### What's this PR do?
Previously, open shift recurrence only checked to see if the `correctNumberOfOpenShiftsToAdd` param was greater than zero. In that case, we'd add shifts. In other cases, we'd do nothing. 

This PR changes that logic so that we check to see if the `correctNumberOfOpenShiftsToAdd` param is negative. (That means that the number of shifts open is greater than what we actually want.) 

In that case, we delete shifts. 

#### how to test this?
Creating open shift times that are 1) greater than the expected amount, 2) less than the expected amount, and 3) just right in the test WiW environment. Running it.

### DIFF
--- a/jobs/scheduling/RecurOpenShifts.js
+++ b/jobs/scheduling/RecurOpenShifts.js
@@ -28,21 +28,21 @@ function recurOpenShifts(now) {
         // Because we're making async calls in a loop, we pass targetTime through callback
         // so that it's defined locally for each scope.
         var targetTime = now.clone().add(i * -2, 'hours');
-        findIfOpenShiftsNeedToBeAddedAndOpenShiftIDsAndOccupiedShiftCount(targetTime, function(openShiftIDs, correctNumberOfOpenShiftsToAdd, targetTime) {
+        findExtraOpenShiftsToDeleteAndOccupiedShiftCount(targetTime, function(extraOpenShiftsToDelete, correctNumberOfShiftsToSet, targetTime) {
             var batchPayload = [];
 
             // If we don't need to add any new open shifts, we return.
-            if (correctNumberOfOpenShiftsToAdd === 0) {
+            if (correctNumberOfShiftsToSet === 0) {
                 console.log('No open shifts need to be added for time: ', targetTime.toString());
                 return;
             }
             // If we need to add open shifts
-            else if (correctNumberOfOpenShiftsToAdd > 0) {
+            else if (correctNumberOfShiftsToSet > 0) {
                 var newOpenShiftParams = {
                     start_time: targetTime.clone().minute(0).second(0).add(1, 'week').format(WIWDateFormat),
                     end_time: targetTime.clone().minute(0).second(0).add(2, 'hour').add(1, 'week').format(WIWDateFormat),
                     location_id: global.config.locationID.regular_shifts,
-                    instances: correctNumberOfOpenShiftsToAdd,
+                    instances: correctNumberOfShiftsToSet,
                     published: true
                 };
                 newOpenShiftParams = colorizeShift(newOpenShiftParams);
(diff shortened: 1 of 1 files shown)


jobs/scheduling/RecurOpenShifts.js
```diff
@@ -52,26 +53,27 @@ function recurOpenShifts(now) {
...
-                    };
-                    batchPayload.push(shiftDeleteRequest);
-                });
-
-                console.log('Adding ', correctNumberOfOpenShiftsToAdd, ' open shifts to the time: ', targetTime.toString(), '. Deleting incorrect count of open shifts--their shift IDs: ', openShiftIDs);
-                WhenIWork.post('batch', batchPayload);
             }
+            // Batch delete the invalid open shifts in two cases: 1) correctNumberOfOpenShiftsToAdd > 0; we need to add open shifts
+            // and 2) correctNumberOfOpenShiftsToAdd < 0; our open shift count is too high
+            openShiftIDs.forEach(function(shiftID) {
```
cjcodes: So at this point, openShiftIDs = [] if there are no openshifts to delete, right? Otherwise, this would delete all openshifts.

cjcodes: For clarity--it may be worth putting this in the block above--it really only needs to run the forEach if there are shifts to add. Right?

jobs/scheduling/RecurOpenShifts.js
```diff
@@ -52,26 +53,27 @@ function recurOpenShifts(now) {
...
+            if (correctNumberOfOpenShiftsToAdd < 0) { correctNumberOfOpenShiftsToAdd = 'no'; }
+            console.log('Adding ', correctNumberOfOpenShiftsToAdd, ' open shifts to the time: ', targetTime.toString(), '. Deleting incorrect count of open shifts--their shift IDs: ', openShiftIDs);
+            WhenIWork.post('batch', batchPayload);
         })
     }
 }
 
 // Checks if open shifts are already present a week from the targetTime
-// Callback params: callback(openShiftsNeedToBeAddedBOOL, openShiftIDs, correctNumberOfOpenShiftsToAdd, targetTimeMomentObj);
+// Callback params: callback(openShiftIDs, correctNumberOfOpenShiftsToAdd, targetTimeMomentObj);
```
cjcodes: Can we call the first param of the callback something different? extraOpenShiftsToDelete or something? Took me a while to understand what this was doing.

jobs/scheduling/RecurOpenShifts.js
```diff
@@ -28,15 +28,16 @@ function recurOpenShifts(now) {
...
+            if (correctNumberOfOpenShiftsToAdd === 0) {
                 console.log('No open shifts need to be added for time: ', targetTime.toString());
                 return;
             }
-            // If the correct number of open shifts haven't been added
-            else {
-                var batchPayload = [];
-                // Add the correct number of open shifts
+            // If we need to add open shifts
+            else if (correctNumberOfOpenShiftsToAdd > 0) {
```
cjcodes: Let's set this variable to not just add--I found myself asking "what if the correct number of shifts to add to the existing shift is negative?"

Something more like correctNumberOfShiftsToSet or something?

cjcodes:
A few notes for clarity--took me a while to wrap my head around the changes.

jobs/scheduling/RecurOpenShifts.js
tongxiang: No, because we're going to batch delete all found open shifts if `correctNumberOfShiftsToSet` < 0. (for instance, if there are 40 people signed up for a shift time for which we only want 20 max shifts, we need to delete all open shifts we find for that time since `correctNumberOfShiftsToSet` would be negative.

tongxiang:
Tested successfully 4:16pm on Friday March 18th

cjcodes:
LGTM

PR #58, opened 2016-03-24
Bulk edit all user accounts to include {canonicalEmail: ''} property with their real email address.

cjcodes:


tongxiang:
looks good to me! :+1:

tongxiang:
So this'll require us to have an array of emails present in the `tasks` directory labelled `emails.js` 

1. To avoid committing these emails, we should add the file to the .gitignore
2. Perhaps we should make sure that this file is called `canonicalEmails.js`, to make it explicit that we should be populating the emails array with the `canonical_emails` column in the CTL platform's user table?

PR #63, opened 2016-03-31
Add makeup shift deletion functionality

tongxiang:
#### What's this PR do?
Adds makeup shift deletion functionality. 

Core logic happens in `www/routes/scheduling/index.js`. 

Running `node app.js` on your local box without passing any `NODE_ENV` env variables will make this run in the test environment, and the default `regular_shifts` location the test location. 

Within WhenIWork, pick up some shifts from both the makeup shifts and test locations. Then, navigate to 
`http://localhost:3000/scheduling/shifts?email=<email>&token=<token>` and then delete some shifts, homie!

#### Questions:
Currently, API calls take a long time--considering decreasing search field for first API call from 365 days? 

Also--add a spinning wheel between when call is made and when it happens.

### DIFF
--- a/www/routes/scheduling/index.js
+++ b/www/routes/scheduling/index.js
@@ -42,7 +42,7 @@ router.get('/shifts', function(req, res) {
                 var query = {
                     user_id: userID,
                     start: '-1 day',
-                    end: '+365 days',
+                    end: '+180 days',
                     location_id: [global.config.locationID.regular_shifts, global.config.locationID.makeup_and_extra_shifts]
                 };
 
@@ -130,8 +130,6 @@ router.get('/shifts', function(req, res) {
                         userName: userName
                     };
 
-                    console.log(templateData);
-
                     // Then, display them in the jade template.
                     res.render('scheduling/chooseShiftToCancel', templateData);
                 })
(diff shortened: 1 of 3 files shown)


www/routes/scheduling/index.js
```diff
@@ -38,8 +42,8 @@ router.get('/shifts', function(req, res) {
                 var query = {
                     user_id: userID,
                     start: '-1 day',
-                    end: '+8 days',
-                    location_id: global.config.locationID.regular_shifts
+                    end: '+365 days',
```
tongxiang: can this be safely brought down to a more reasonable number so that we're not querying for a bajillion shifts?

tongxiang: alternately, we can make two API calls--one for regular shifts and one for makeup shifts. This may take a shorter amount of time.

tongxiang: the API call is taking a really long time for me because I have signed up for so many test shifts for my account.

www/routes/scheduling/index.js
```diff
@@ -1,18 +1,22 @@
...
-
-var router = express.Router();
-
-var api = new WhenIWork(global.config.wheniwork.api_key, global.config.wheniwork.username, global.config.wheniwork.password);
-
-var wiwDateFormat = 'ddd, DD MMM YYYY HH:mm:ss ZZ';
-var chooseShiftToCancelPageStartDateFormat = 'dddd h:mm a';
-var chooseShiftToCancelPageEndDateFormat = 'h:mm a z';
-var scheduleShiftsURL = '/scheduling/login?';
+var express   = require.main.require('express')
```
cjcodes: Technically speaking, there is no performance optimization in using fewer var statements. Rather, this is a preference, so I'd like to have a consistent coding preference that we adhere to. If multi-line commas is how we're gonna do it, let's stick with it throughout.

My opinion is that non-comma style is easier to maintain, and the readability gain is marginal. But that's just my opinion. :P

cjcodes:
:+1:

PR #66, opened 2016-04-03
colorize makeup shifts created by timeoff requests, refactor `colorizeShift` accordingly

tongxiang:
#### What's this PR do?
Makeup shifts created by timeoff requests were not deliberately being colorized. This colorizes them to be light gray if they're non-urgent, and red if they are during urgent times. 

This also changes how `colorizeShift` works to support makeup shifts being colorized a different way. 

Tested by confirming that makeup shifts.

### DIFF
--- a/app.js
+++ b/app.js
@@ -2,7 +2,8 @@ global.config = require('./config');
 require('newrelic');
 
 if (process.env.NODE_ENV !== 'production') {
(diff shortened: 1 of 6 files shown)


tongxiang:
:+1: <-- from @cjcodes

PR #68, opened 2016-04-04
add merge makeup shift functionality, refactor, merge task

tongxiang:
#### What's this PR do?
Previously, makeup shifts weren't getting merged. (There were many makeup shifts appearing for same time as opposed to one makeup shift with many instances.)

This PR updates the job so that they are merged together, and colorizes them properly. 

This also adds a task so we can run the merge ourselves. 

Tested in the `test` and `test2` environments, then tested again by running against production via the task. Works.


cjcodes:
:+1:

PR #71, opened 2016-04-06
Add an option for users to set their timezone if they haven't yet set it.

cjcodes:


www/routes/scheduling/index.js
```diff
@@ -309,6 +310,28 @@ router.get('/login', function (req, res) {
                 res.redirect('https://app.wheniwork.com');
                 return;
             } else {
+                // If they have not yet set their timezone
+                if ((user.notes.indexOf('timezoneSet') < 0) && req.query.timezone == undefined) {
```
tongxiang: I think it'd be best to `JSON.parse(user.notes)` first before checking for the existence of this param.

tongxiang: Have we tested this for users who already have a `timezoneSet` param in their notes?

www/routes/scheduling/index.js
```diff
@@ -309,6 +310,28 @@ router.get('/login', function (req, res) {
...
+                    res.redirect('/scheduling/timezone?' + querystring.stringify(req.query));
+                    return;
+                }
+                // If they are coming via the timezone route (they've selected a timezone)
+                // Note this is all done in the background
+                else if (req.query.timezone !== undefined && req.query.timezone !== '') {
+                    // Parse notes
+                    var notes = {};
+                    if (user.notes.indexOf('{') >= 0) {
```
tongxiang: What about users who have notes that aren't a JSON object? Do we have any users who still have notes that are a pure string? *Will we?*

www/routes/scheduling/index.js
```diff
@@ -309,6 +310,28 @@ router.get('/login', function (req, res) {
...
+                        notes = JSON.parse(user.notes);
```
tongxiang: I think we should wrap all our `JSON.parse` calls in a try/catch.

www/routes/scheduling/index.js
```diff
@@ -321,6 +344,28 @@ router.get('/login', function (req, res) {
...
+        13: 'Mountain',
+        170: 'Arizona',
+        15: 'Pacific',
+        19: 'Hawaii',
+        167: 'Alaska'
+    };
+
+    var url = '/scheduling/login';
+
+    res.render('scheduling/timezone', {url: url, params: req.query, timezones: timezones});
```
tongxiang: Why do we want to limit their timezone selection to just US timezones?

tongxiang:
:+1:

PR #80, opened 2016-05-14
refactor scheduling/shifts routes

tongxiang:
#### What's this PR do?
Previously, the `scheduling/index.js` route was over 500 lines long. Difficult to read, and difficult to test. 

This PR is a first attempt at modularizing the `scheduling` routes. 

1. It modularizes the `scheduling/shifts` route into a small Express router which is then required by `scheduling.index.js`
2. It creates a `helpers.js` file that's used by both `scheduling/index.js` and `scheduling/shifts/index.js`
3. It modifies the route initialization call in `app.js` to be a bit clearer. 
4. Because we're now requiring use of the same WhenIWork API in two places, I created a `initWhenIWorkAPI.js`, which initializes the primary WhenIWork API object once. It's then required in both `scheduling/index.js` and `scheduling/shifts/index.js`. (Otherwise, we'd require two separate whenIWork API invocations.)
5. Refactors the directory structure to be more MVC-y. (Moves `scheduling` out of `routes` into its own folder, in preparation for adding a `controllers` directory. Then, deletes `routes` directory.) 
6. Modularizes route logic into controllers which can be tested without hitting these routes. 
7. Adds a test for `retrieveAndRendersShiftsToDelete`, for main shift deletion view. 

Start by seeing how the route initialization in `app.js` changes. 
Then, see what's been removed in `www/routes/scheduling/index.js` and moved to `www/routes/scheduling/shifts/index.js`. 

Tested by running on my local machine and running through all the routes moved to `scheduling/shifts/index.js`--the entirety of the shift deletion flow. Then the `scheduling/shifts/time-interval` route was tested with Postman. 

This was done in the service of modularizing Elmer so that the `/scheduling` and `/scheduling/shifts` routes can be more easily testable. 

Because I couldn't figure out a clean way of testing the generated views spit out by the routes, my plan is to add event emitters to the controller functions, emitting data at critical times (for instance, template data), and then my tests will check for the presence/absence of values in that data. Not ideal, but the best solution, I think. 

https://admin.crisistextline.org/jira/browse/INT-87

#### Questions:
Any criticism of this refactoring / testing plan, @cjcodes?


test/retrieveAndRenderShiftsToDelete.js
```diff
@@ -0,0 +1,66 @@
+var nock = require('nock')
+  , WhenIWork = require('wheniwork-unofficial')
+  , config = require('../config')
```
cjcodes: Is this a strong dependency? Shouldn't we be mocking out the config stuff too?

test/retrieveAndRenderShiftsToDelete.js
```diff
@@ -0,0 +1,66 @@
...
+  ;
+
+describe('retrieveAndRenderShiftsToDelete should retrieve shifts and render them when user navigates to shift deletion page', function() {
+  var base = nock('https://api.wheniwork.com/2');
+  var loginResponse = {
+    login: {
+      token: 'FAKETOKEN'
+    }
+  };
+  var userEmail = 'amudantest@gmail.com';
```
cjcodes: Was this "a mundane test" ? Lol. Will this poor sap get emails ever?

test/typicalUserResponse.js
```diff
@@ -0,0 +1,54 @@
...
+  account_id: 549781,
+  login_id: 2761909,
+  timezone_id: 0,
+  country_id: 233,
+  migration_id: 0,
+  role: 3,
+  is_payroll: false,
+  is_trusted: 0,
+  type: 1,
+  email: 'ambuncher27@gmail.com',
```
cjcodes: Is this real user data? Can we make it not if it is?

www/scheduling/index.js
```diff
@@ -0,0 +1,155 @@
...
+    };
+
+    var url = '/scheduling/login';
+
+    res.render('scheduling/timezone', {url: url, params: req.query, timezones: timezones});
+});
+
+
+function checkUser(email, first, last, callback) {
+    var altEmail = email.replace(/\W+/g, '');
```
cjcodes: These two lines should probably be in a module / helper somewhere (`generateAltEmail(email)`) since I think we'll reuse this a lot.

www/scheduling/initWhenIWorkAPI.js
```diff
@@ -0,0 +1,14 @@
+var WhenIWork = require.main.require('wheniwork-unofficial')
+  , api
+  ;
+
+function initAPI() {
```
cjcodes: Any reason for doing it this way? An alternative could be to use a var and initialize it immediately on import. So:
```
var api = new WhenIWork(...);

module.exports = api;
```

Or you could one-line it:
```
module.exports = new WhenIWork(...);
```

Because node will re-use a module once you import it once.

cjcodes:
Couple comments. Let me know when resolved.

test/retrieveAndRenderShiftsToDelete.js
tongxiang: I don't think it's necessary to mock out config. 

Having a separate `testConfig` file would mean that we'll need to make config updates in two places. 

A lot of stuff in `config` is necessary for testing--the color of shifts, location IDs. 

Open for your opinion on this, though. I don't know the best way we'd do this.

test/retrieveAndRenderShiftsToDelete.js
tongxiang: Changed this to be a fake(r) email.

test/typicalUserResponse.js
tongxiang: This file was unnecessary, so I deleted it.

www/scheduling/index.js
tongxiang: Done.

cjcodes:
:+1:

PR #91, opened 2016-06-01
Config

mbogor:
#### What's this PR do?
refactors the config file into production and test keys file and the regular config file
keys.js.default README config.js.default
follow instructions in the readme; more thorough tests will be available after i add the keys and keys_test invocations throughout the elmer directory
not at the moment
'remove invocation of production config file from tests'
#### Questions:
why do banks charge minimum balance fees when they know you don't got no money?


README.md
```diff
@@ -8,7 +8,9 @@ Elmer, like da glue, holds everything together. Currently, it extends the functi
 5. deleting shifts
 
 ## Running on local
-Run `npm install`. In your main directory, `touch config.js`. Copy-and-paste the contents of `config.js.default` into this new config file. Add requisite API keys.
+Run `npm install`. In your main directory, `touch config.js`. Copy-and-paste the contents of `config.js.default` into this new config file.
+
+In your main directory, `touch keys.js keys_test.js`. Copy-and-paste the contents of `keys.js.default` into the new key files. Add requisite API keys.
```
tongxiang: Very thorough--thinking to update the `README.md`.

tongxiang:
If we're extracting all the non-commitable keys out of the config file, we no longer need the separation between `config.default.js` and `config.js`. 

1) Let's just move everything in a single `config.js` file and commit that. 

2) Since merging in this PR would break the app, let's add commits to this PR integrating the new `config.js` and `keys.js` file into the app. Let's make sure all tests run, and the app runs, before merging in.

PR #101, opened 2016-06-07
Refactored recurOpenShifts tests and RecurOpenShifts production file …

jrauschenberg:
#### What's this PR do?
It changes the recurOpenShifts tests and the RecurOpenShifts production file so that the tests run on the actual production file, rather than a duplicate file created just for testing purposes. I also added a failure case test and a test for a new function, incrementFutureOpenShiftsUpOrDown, that was put into RecurOpenShifts.
test/recurOpenShifts and jobs/scheduling/RecurOpenShifts.
"npm test"
[INT-81](https://admin.crisistextline.org/jira/secure/RapidBoard.jspa?rapidView=5&view=detail&selectedIssue=INT-81)
#### Questions:

…so that a separate test file is not used.


jobs/scheduling/RecurOpenShifts.js
```diff
@@ -15,30 +15,39 @@ function runJob() {
...
-// run it once on boot
-runJob();
+if (process.env.NODE_ENV !== 'test') {
+  // run it once on boot
+  runJob();
+}
 
 // Recurs open shifts one and two weeks in the future.
-function recurOpenShifts(now) {
+function recurOpenShifts(now, sampleData) {
```
tongxiang: We should rename this so that it's clear that it's only used in the test environment. Like: `optionalSampleDataUsedInTesting`.

jobs/scheduling/RecurOpenShifts.js
```diff
@@ -15,30 +15,39 @@ function runJob() {
...
     // Because we're making async calls in a loop, we pass targetTime through callback
     // so that it's defined locally for each scope.
     var targetTime = now.clone().add(i * -2, 'hours');
-    findExtraOpenShiftsToDeleteAndOccupiedShiftCount(targetTime, 1, incrementFutureOpenShiftsUpOrDown);
-    findExtraOpenShiftsToDeleteAndOccupiedShiftCount(targetTime, 2, incrementFutureOpenShiftsUpOrDown);
+
+    if (process.env.NODE_ENV === 'test') {
+      return findExtraOpenShiftsToDeleteAndOccupiedShiftCount(targetTime, 2, incrementFutureOpenShiftsUpOrDown, sampleData);
+    }
+    findExtraOpenShiftsToDeleteAndOccupiedShiftCount(targetTime, 1, incrementFutureOpenShiftsUpOrDown, sampleData);
```
tongxiang: Why are we passing in `sampleData` in when this runs on production?

jobs/scheduling/RecurOpenShifts.js
```diff
@@ -58,30 +67,36 @@ function findExtraOpenShiftsToDeleteAndOccupiedShiftCount(targetTimeMomentObj, w
...
-      if (JSON.parse(shift.is_open)) {
-        if (!shift.instances) {
-          countOfOpenShifts++;
-        }
-        else {
-          countOfOpenShifts += JSON.parse(shift.instances);
-        }
-        extraOpenShiftsToDelete.push(shift.id);
+  if (process.env.NODE_ENV === 'test') {
+    return countOpenShifts(sampleData.shifts, countOfOpenShifts, countOfOccupiedShifts, weeksFromNowToCheck, targetTimeMomentObj, extraOpenShiftsToDelete, callback);
```
tongxiang: Instead of returning a new function if the NODE_ENV === test, let's just have this function `findExtraOpenShiftsToDeleteAndOccupiedShiftCount` return the `filter` that's passed into the get request. 

Then, let's individually test the `findExtraOpenShiftsToDeleteAndOccupiedShiftCount` and `countOpenShifts` functions in the test file by exporting them. 

That way, we can cut down on returning functions in extra, test-specific control flow. (This is a bit messy.)

tongxiang:
For some reason, tests are not passing on this branch?

PR #104, opened 2016-06-08
Refactored login and user creation tests to use the original producti…

jrauschenberg:
#### What's this PR do?
It refactors the "loginusercreation" tests to use the production file in www/scheduling/index.js, instead of a completely separate file used just for testing. It also refactors the index.js file to extract the main functionality from the "checkUser" function into a separate function that can be used by both production and the test file.
www/scheduling/index.js and test/loginusercreation.js.
"npm test" or by manually logging in or creating a new user in WiW.
Tickets either never existed or have been deleted.
#### Questions:

…on index.js file, rather than a separate file just for testing. Extracted functionality from the checkUser function in the index.js file to allow for testing.


tongxiang:
This looks like it'd work. Have you been able to run this on your local and test against WhenIWork with Postman and hashing your email with the secret? If not, make sure you test 1) that you're able to login with a pre-existing user, and 2) that you're able to login and create a new user, if one doesn't already exist. Then, go into WiW and confirm that that user has been created.

PR #105, opened 2016-06-08
fixed JSON.parse bug

mbogor:
#### What's this PR do?
fixes JSON.parse error in delete shift and redirect file
www/scheduling/shifts/controllers/deleteShiftsAndRedirect.js
'debug JSON.parse() call in deleteShifts view'


www/scheduling/shifts/controllers/deleteShiftsAndRedirect.js
```diff
@@ -40,8 +40,9 @@ function deleteShiftsAndRedirect(req, res, whenIWorkAPI) {
     var shift;
     var batchPayload = [];
     var deletedShiftInformation = {regShifts : {}, makShifts : {}};
+    var shifts = data.shifts[0].shifts;
```
tongxiang: the data format returned by the WIW API in the response body is: 

`{shifts: [array_of_shifts] }` 

Confused how this is a solution for the bug? 

Has this been tested on production?

PR #112, opened 2016-06-14
Wiw mocker

jrauschenberg:
#### What's this PR do?
Refactors NotifyFirstShift to use a WiW mocker (in the config file) in test mode.
#### Where should the reviewer start?y
config.js, jobs/scheduling/NotifyFirstShift, test/notifyFirstShift
"Npm test" or using When I Work.
[INT-79](https://admin.crisistextline.org/jira/browse/INT-79)


apackin:
👍

apackin:
👍 @tongxiang

jobs/scheduling/NotifyFirstShift.js
```diff
@@ -11,64 +10,76 @@ new CronJob(CONFIG.time_interval.notify_first_shift_cron_job_string, function ()
...
           }
-        });
-      }
+      };
+      mandrill_client.messages.send({message: message, key: 'first_shift_scheduled'}, function (res) {
+          CONSOLE_WITH_TIME(res);
+      });
+
+      WhenIWork.update('users/' + user.id, {notes: user.notes + 'first_shift_notified' + "\n"});
+      return ({user: user.id, email: user.email, name: user.first_name + " " + user.last_name, notes: user.notes, message: message});
```
tongxiang: Why are we returning within this `for` loop? Doesn't that prevent the email from being sent to any other users than the first one?

jobs/scheduling/base.js
```diff
@@ -1,3 +1,6 @@
+var CONFIG = require('../../config.js');
+var KEYS = require('../../keys.js');
```
tongxiang: Is there a way we can invoke this `base.js` later so that we don't need to require `CONFIG` and `KEYS` again in this file? 

Since we've already declared `CONFIG` and `KEYS` as global variables, it'd be good to not have to declare these again locally. 

Potential solutions: 1) move the base invocation to a place where `KEYS` are already loaded? 

2) make the `CONFIG.WhenIWork` invocation a function, that's run after `KEYS` is loaded.

jobs/scheduling/updateCanvas.js
```diff
@@ -0,0 +1,23 @@
+function updateCanvas(user, course, assignment) {
```
tongxiang: Since 1) this isn't functioning and 2) this isn't relevant to the subject matter of this commit, let's take this file out and add it to the next PR.

test/sampleData.js
```diff
@@ -588,8 +592,8 @@ var wiwUsersObject = [
     last_login: 'Mon, 25 Jan 2016 21:30:38 -0500',
     dismissed_at: '',
     notified_at: 'Tue, 29 Nov -001 19:00:00 -0500',
-    created_at: 'Tue, 19 Jan 2016 14:48:09 -0500',
-    updated_at: 'Mon, 25 Jan 2016 21:32:47 -0500',
+    created_at: moment().subtract(1, 'weeks'),
```
tongxiang: Add a comment as to why you're adding this moment invocation

test/timeOffRequests.js
```diff
@@ -4,7 +4,7 @@ var assert = require('assert'),
 
 var requests = sampleData.requestsResponse.requests;
 var shiftsResponse = {
-  shifts: sampleData.shifts.shifts.slice(0,4)
+  shifts: sampleData.shiftsResponse.shifts.slice(0,4)
```
tongxiang: What is this fixing? IF it's fixing something, note it in the PR

jobs/scheduling/NotifyFirstShift.js
```diff
@@ -11,64 +10,76 @@ new CronJob(CONFIG.time_interval.notify_first_shift_cron_job_string, function ()
   checkNewShifts();
 }, null, true);
 
-checkNewShifts();
```
tongxiang: This is here so that the function runs on boot. 

I'm agnostic as to whether or not we want this to run on boot, but for the sake of consistency I think we should keep it in.

test/timeOffRequests.js
apackin: We changed the format of the sampleData to be more consistent with the responses from requests. This is accounting for that change.

jobs/scheduling/NotifyFirstShift.js
apackin: This return only applies to getShifts() - not a return out of the forEach

apackin: I realized your referring to ' for (var j in shifts)' ... 
This does prevent the email from being sent out for more than one shift, but it's only meant to be sent for their first shift anyway, therefore I'm not sure why the loop is necessary.

test/sampleData.js
jrauschenberg: Done!

jobs/scheduling/updateCanvas.js
jrauschenberg: Done!

jobs/scheduling/base.js
jrauschenberg: Assaf, do you want to pair up to come up with a solution to this?

jobs/scheduling/NotifyFirstShift.js
jrauschenberg: Put it back in there.

www/app.js
```diff
@@ -1,5 +1,6 @@
-global.CONFIG = require('../config');
```
jrauschenberg: We changed the order of requirement of keys and config here, since config calls .base, which depends on keys. We think that by requiring them in this order here, we won't need to call them later in the jobs/scheduling/.base file. We're not sure how to test this yet. Please let us know how we might do it. Thanks. @tongxiang

tongxiang:
feel free to merge after testing the config / keys arrangement @apackin

PR #114, opened 2016-06-16
Recur shifts refactored

jrauschenberg:
#### What's this PR do?
Refactors recurShifts and the recurShifts tests to take advantage of the Wiw mocker.
test/recurShifts and jobs/scheduling/RecurShifts.js
"npm test" or run RecurShifts in node.
[INT-82
](https://admin.crisistextline.org/jira/secure/RapidBoard.jspa?rapidView=5&projectKey=INT&view=detail&selectedIssue=INT-82)#### Questions:


jobs/scheduling/RecurShifts.js
```diff
@@ -10,10 +10,11 @@ new CronJob(CONFIG.time_interval.recur_and_publish_shifts_cron_job_string, funct
   recurNewlyCreatedShifts();
 }, null, true);
 
-recurNewlyCreatedShifts();
-
-function recurNewlyCreatedShifts() {
+function recurNewlyCreatedShifts(optionalNoShiftsForTesting) {
```
apackin: Can we think of a way to test the same thing without adding params that are only for testing?

jrauschenberg: Yeah I thought about this for a while and couldn't find anything more elegant. The problem is without this param, our base will give the function a bunch of sample data. I guess we could refactor the base somehow to allow it to return blank sampleData if we want? Let me know if you have ideas.

jobs/scheduling/RecurShifts.js
```diff
@@ -10,10 +10,11 @@ new CronJob(CONFIG.time_interval.recur_and_publish_shifts_cron_job_string, funct
   recurNewlyCreatedShifts();
 }, null, true);
 
-recurNewlyCreatedShifts();
```
tongxiang: The point of this function call was to run this job on start. 

Was there a reason you removed it?

tongxiang:
Tests seem to be failing on this branch, even when I comment out all the other tests in the `test/index.js` file.

jobs/scheduling/RecurShifts.js
tongxiang: I think I would prefer to not test a path, than to add an optional parameter that's potentially confusing. 

(Because basically what you're trying to test is that if no shifts are returned, we have no payload.)

My preference: to get rid of this param, and the test.

jobs/scheduling/RecurShifts.js
```diff
@@ -192,34 +193,33 @@ function recurNewlyCreatedShifts() {
...
       async.waterfall(requestTaskArray, function(err, startDate, endDate, unpublishedShiftIDs) {
-        var publishPayload = {
+        publishPayload = {
           'ids': unpublishedShiftIDs
         };
-
         WhenIWork.post('shifts/publish/', publishPayload);
       });
     });
   });
```
tongxiang: I'd make a note here that this function is returning *specifically* for testing.

jobs/scheduling/RecurShifts.js
```diff
@@ -28,7 +29,7 @@ function recurNewlyCreatedShifts() {
                   };
 
   WhenIWork.get('shifts', postData, function(response) {
-    var allShifts = response.shifts;
+    var allShifts = optionalNoShiftsForTesting || response.shifts;
     if (typeof allShifts !== 'object') {
```
apackin: If the purpose of optionalNoShiftsForTesting is to test this error handling, can we extract it to its own function and test it independently ?

jobs/scheduling/RecurShifts.js
```diff
@@ -149,6 +148,7 @@ function recurNewlyCreatedShifts() {
             });
 
             unpublishedShiftIDs = unpublishedShiftIDs.concat(newUnpublishedShiftIDs);
+            globalUnpublishedShiftIDs = unpublishedShiftIDs;
```
tongxiang: 1. I wouldn't use the word `global` to describe this variable, since it's not actually a global. 
2. What is this variable being used for? It's not used in our tests. 
3. Note that `task` is run multiple times after `firstTask` is run once. Therefore, this variable is getting overwritten multiple times.

tongxiang: Nope, I'm wrong--this is getting properly updated.

tongxiang:
👍

PR #116, opened 2016-06-16
Recur open shifts refactored

jrauschenberg:
#### What's this PR do?
It's version 3 of working up tests for RecurOpenShifts. In addition to refactoring the tests, it changes the RecurOpenShifts production file slightly to take advantage of the new WiW mocker.
test/recurOpenShifts.js and jobs/scheduling/RecurOpenShifts.js.
"npm test" or run jobs/scheduling/RecurOpenShifts in node.
[INT-81](https://admin.crisistextline.org/jira/secure/RapidBoard.jspa?rapidView=5&projectKey=INT&view=detail&selectedIssue=INT-81)


apackin:
👍

jobs/scheduling/RecurOpenShifts.js
```diff
@@ -130,6 +132,7 @@ function incrementFutureOpenShiftsUpOrDown(extraOpenShiftsToDelete, correctNumbe
   if (correctNumberOfShiftsToSet < 0) { correctNumberOfShiftsToSet = 'no'; }
   CONSOLE_WITH_TIME('Adding ', correctNumberOfShiftsToSet, ' open shifts to the time: ', targetTime.toString(), ' . Deleting incorrect count of open shifts--their shift IDs: ', extraOpenShiftsToDelete);
   WhenIWork.post('batch', batchPayload);
+  return batchPayload;
```
tongxiang: Would add another note about how you're returning values for test purposes here, too.

jobs/scheduling/RecurOpenShifts.js
```diff
@@ -78,8 +79,9 @@ function findExtraOpenShiftsToDeleteAndOccupiedShiftCount(targetTimeMomentObj, w
     var correctNumberOfShiftsToSet = returnMaxOpenShiftCountForTime(targetTimeMomentObj.clone()) - countOfOccupiedShifts;
     CONSOLE_WITH_TIME('Found ', countOfOpenShifts, ' open shifts, ', countOfOccupiedShifts, ' occupied shifts found for a time ', weeksFromNowToCheck, 'weeks from now where we expect ', returnMaxOpenShiftCountForTime(targetTimeMomentObj.clone()), ' open shifts. Time: ', targetTimeMomentObj.toString());
     callback(extraOpenShiftsToDelete, correctNumberOfShiftsToSet, weeksFromNowToCheck, targetTimeMomentObj);
-    return;
   });
+  //Just returning values for test purposes here.
+  return [countOfOpenShifts, countOfOccupiedShifts, extraOpenShiftsToDelete];
```
tongxiang: If this `return` happens outside of the `WhenIWork.get` callback, isn't the values it's returning not being updated?

tongxiang: Ahhh. This works during testing because the mocker makes it run synchonously. 

This is confusing, though--this should work even if the code doesn't run synchronously. 

This doesn't seem to work if we move the return statement inside the callback--why?

test/recurOpenShifts.js
```diff
@@ -0,0 +1,37 @@
...
+var oddHourTargetTime = moment("2016-06-03T11:00:00.176Z");
+var recurOpenShifts = require('../jobs/scheduling/RecurOpenShifts.js');
+
+describe('recurOpenShifts', function() {
+
+    it('should recur open shifts', function (done) {
+      var result = recurOpenShifts.recurOpenShifts(targetTime);
+      assert.equal(result[0], 5);
+      assert.equal(result[1], 6);
+      assert.equal(result[2][0], 284948049);
```
tongxiang: It's not easy to understand what each one of the `assert.equal` statements is checking for in the test. 

My suggestion:

Move the `recurOpenShifts.recurOpenShifts` call into a broader `describe` statement, and then have each `assert.equals` within a `it("should do something", function(done)` statement that clearly expresses what the value we're checking for is.

tongxiang: That way, we know whether or not it's `countOfOpenShifts`, `countOfOccupiedShifts`, or `extraOpenShiftsToDelete` that's breaking our test.

tongxiang:
made some comments @jrauschenberg!

jobs/scheduling/RecurOpenShifts.js
jrauschenberg: @tongxiang it doesn't work because it only returns out of the callback, not the overall findExtraOpenShiftsToDeleteAndOccupiedShiftCount function. What change do you suggest I make? Thanks!

jrauschenberg:
@tongxiang I fixed everything other than as noted above (with the placement of the return statement). Let me know how to proceed--thanks.

jrauschenberg:
Made the changes we discussed, including incorporating an event emitter, @tongxiang. Should be ready to merge!

jobs/scheduling/RecurOpenShifts.js
```diff
@@ -77,9 +83,13 @@ function findExtraOpenShiftsToDeleteAndOccupiedShiftCount(targetTimeMomentObj, w
 
     var correctNumberOfShiftsToSet = returnMaxOpenShiftCountForTime(targetTimeMomentObj.clone()) - countOfOccupiedShifts;
     CONSOLE_WITH_TIME('Found ', countOfOpenShifts, ' open shifts, ', countOfOccupiedShifts, ' occupied shifts found for a time ', weeksFromNowToCheck, 'weeks from now where we expect ', returnMaxOpenShiftCountForTime(targetTimeMomentObj.clone()), ' open shifts. Time: ', targetTimeMomentObj.toString());
+    //the below just runs for test purposes.
+    if (process.env.NODE_ENV === 'test') {
+      eventEmitter.emit('testData', {countOfOpenShifts: countOfOpenShifts, countOfOccupiedShifts: countOfOccupiedShifts, extraOpenShiftsToDelete: extraOpenShiftsToDelete});
```
tongxiang: Why are you emitting this string if you're defining `{testData: 'testData'}` above? I'd choose one or the other.

jobs/scheduling/RecurOpenShifts.js
tongxiang: Bump ^

jobs/scheduling/RecurOpenShifts.js
tongxiang: Personally, I'd just pass a string, like `testData`. No need to define a separate object until you need to.

test/recurOpenShifts.js
```diff
@@ -0,0 +1,60 @@
...
+var eventEmitter = require('../jobs/scheduling/RecurOpenShifts.js').eventEmitter;
+var result;
+
+describe('recurOpenShifts', function() {
+
+    it('should accurately find the number of open shifts', function (done) {
+      eventEmitter.on('testData', testData);
+      function testData(data) {
+        result = data;
+        assert.equal(result.countOfOpenShifts, 6);
```
tongxiang: After you the test `assert.equal` call runs, remove the eventEmitter. Like so: 

`emitter.removeAllListeners('testData');`

Otherwise the functions will continue to run after `testData` gets emitted again. 

Also, it'd be a good idea to rename `testData` so that it's a string that's more specific to this test's use-case. (So that other event-driven tests don't collide with that namespace.)

apackin:
👍

tongxiang:
@jrauschenberg 

I remember that we identified a problem with how this code was running [with this JIRA ticket.](https://admin.crisistextline.org/jira/browse/INT-218)

Was there a bug? If so, how does this code solve that bug? How did you debug it?

jrauschenberg:
@tongxiang It wasn't so much a bug with the code as in how we were testing it. We were looking for it to run on the test WiW location, but weren't setting it to do so. Once I did that, it recurred shifts perfectly on the test location in a production environment.

tongxiang:
ahhh. So the config changes if we're not running prod happen in `runJobs.js`, and weren't testing by running that file. Makes sense.

tongxiang:
👍  

Great job, John.

PR #117, opened 2016-06-17
Refactored loginusercreation tests and slightly modified www/scheduli…

jrauschenberg:
#### What's this PR do?
Refactors the loginusercreation tests and slightly modified www/scheduling/index.js to use the new WiW mocker.
tests/loginusercreation and www/scheduling/index.js.
"Npm test" (but only if you have the updates to test/helpers/base.js and www/scheduling/initWhenIWorkAPI.js already installed. Otherwise the tests will fail. Manually with Node.
None any more, but this is v3 of fixing the loginusercreation tests.
#### Questions:

…ng/index.js to use Wiw mocker.


mbogor:
👍

www/scheduling/initWhenIWorkAPI.js
```diff
@@ -2,4 +2,5 @@ var WhenIWork = require.main.require('wheniwork-unofficial');
 
 // Node will re-use a module once you import it once, so this API will not initialized twice.
 var api = process.env.NODE_ENV === 'test' ? require('../../test/helpers/base') : new WhenIWork(KEYS.wheniwork.api_key, KEYS.wheniwork.username, KEYS.wheniwork.password);
+
```
tongxiang: Let's take this out.

test/index.js
```diff
@@ -6,7 +6,7 @@ console.log('index file running');
 global.KEYS = require('../keys.js');
 global.CONFIG = require('../config.js');
 //tests
-var colorize = require('./colorize.js');
+// var colorize = require('./colorize.js');
```
tongxiang: stray comment

www/scheduling/index.js
```diff
@@ -1,6 +1,7 @@
 var express   = require('express')
-  , WhenIWork = require('wheniwork-unofficial')
+  , WhenIWork = CONFIG.WhenIWork
   , api = require('./initWhenIWorkAPI')
+  , api2 = require('./initWhenIWorkAPI')
```
tongxiang: When `require` is first called, it runs the file and then caches it. The next time it's called, it returns whatever it cached. 

Therefore, this `api2` invocation will likely return the same API, authenticated the same way. 

Let's test this, and figure out a new way to add a new instance of this API.

www/scheduling/index.js
```diff
@@ -151,6 +151,10 @@ function checkUser(email, first, last, callback) {
       });
     });
   });
+  return newUser;
```
tongxiang: Note that this is used specifically for testing

www/scheduling/initWhenIWorkAPI.js
jrauschenberg: @tongxiang What are you referring to? We need the api variable. The comment?

www/scheduling/index.js
jrauschenberg: @tongxiang Ok, cool. Can we team up briefly on this? I'm not sure what the best practice here would be.

tongxiang: Yup! Let's work together on this one.

www/scheduling/initWhenIWorkAPI.js
tongxiang: The extra line break

tongxiang: Let's not disturb the git commit history for this file if no code has been added

www/scheduling/initSecondWhenIWorkAPIwithParams.js
```diff
@@ -0,0 +1,15 @@
...
+var api2;
+
+// Node will re-use a module once you import it once, so this API will not initialized twice.
+var createSecondAPI = function(WiWApiKey, email, WiWDefaultPassword) {
+  if (process.env.NODE_ENV === 'test') {
+    api2 = require('../../test/helpers/base');
+    return api2;
+  } else {
+    api2 = new WhenIWork(WiWApiKey, email, WiWDefaultPassword);
+    return api2;
```
tongxiang: Since this is a function that can be reused, I'd move the `var api2;` API instantiation into the `createSecondAPI` function. 

and then have one `return api2` statement, instead of returning in both of the control flows.

www/scheduling/index.js
```diff
@@ -133,9 +135,8 @@ function checkUser(email, first, last, callback) {
     };
 
     api.post('users', newUser, function (data) {
-      var api2 = new WhenIWork(KEYS.wheniwork.api_key, altEmail, KEYS.wheniwork.default_password, function (data) {
-      });
-
+      
+      api2 = api2(KEYS.wheniwork.api_key, altEmail, KEYS.wheniwork.default_password, function (data) {});
```
tongxiang: you're re-assigning the `api2` variable to itself--I would call it something else so you're not overwriting the variable.

tongxiang:
👍

PR #120, opened 2016-06-20
test debugged

mbogor:
#### What's this PR do?
debugged and denocked delete shift and redirect test


test/deleteShiftsAndRedirect.js
```diff
@@ -131,11 +112,7 @@ describe('delete shifts and redirect', function() {
...
       };
-
       deleteShiftsAndRedirect(request, response, api);
     });
-
   });
-
-});
-
+});
```
jrauschenberg: I think you could DRY this out a bit, since we're using a very similar request/response object over and over.

jrauschenberg: Also, I'm not sure right now what the difference is between the makeup shift test and the regular shift test. It seems that we're testing that a makeup shift has been changed in both places. I might be wrong, but worth a look.

jrauschenberg: Okay, those two comments are it, @mbogor!

mbogor:
hold on merging; the sample data file is undergoing major changes, after those are merged in, this file needs to be refactored again.

PR #127, opened 2016-06-21
Add a churn remover script

cjcodes:


tasks/removeChurned.js
```diff
@@ -0,0 +1,104 @@
+
+var WhenIWork = require('wheniwork-unofficial');
+var api = new WhenIWork(global.config.wheniwork.api_key, global.config.wheniwork.username, global.config.wheniwork.password);
```
apackin: We've been using 
var api = CONFIG.WhenIWork;

tasks/removeChurned.js
```diff
@@ -0,0 +1,104 @@
...
+
+var churned = require('../churned');
+
+module.exports.go = function () {
+  var userList = [];
+  getUsersToClean().then(function (users) {
+    users.forEach(function (e, i, arr) {
```
apackin: Don't need the arr param on forEach if you're not using it.

tasks/removeChurned.js
```diff
@@ -0,0 +1,104 @@
+
+var WhenIWork = require('wheniwork-unofficial');
+var api = new WhenIWork(global.config.wheniwork.api_key, global.config.wheniwork.username, global.config.wheniwork.password);
+
+var churned = require('../churned');
```
tongxiang: What format does `churned` come in?

tongxiang:
Note that this doesn't delete churned users, just their shifts.

PR #126, opened 2016-06-21
Gtw integration

mbogor:
#### What's this PR do?
automates canvas grading for attending GTW observation

PR #131, opened 2016-06-27
Revert "Gtw integration"

tongxiang:
Reverts CrisisTextLine/Elmer#126

Not for anything structurally wrong, but just to review refactored code.


tongxiang:
note that this reversion was itself reverted, and this no longer appears on master.

PR #133, opened 2016-06-27
add fix for secondary API creation

tongxiang:
#### What's this PR do?
Previously, this call: 

`var secondAPI = api2(KEYS.wheniwork.api_key, altEmail, KEYS.wheniwork.default_password, function (data) {});`

Broke our scheduling flow. This fixes it. 

Also runs `recurShifts` on start. 

Tested by hitting the `/login` URL.


jrauschenberg:
Tested locally on my machine and looked over code. Should be good to go. 👍

PR #134, opened 2016-06-27
create comprehensive time off requests dump

apackin:
#### What's this PR do?
Creates comprehensive time off requests task that dumps data to exported CSV
The flow of the file is written out at the top of the module export, functions are written below in order of execution. 
execute module.exports.timeOffRequests() with or without designating params 
https://admin.crisistextline.org/jira/secure/RapidBoard.jspa?rapidView=5&view=detail&selectedIssue=INT-181
#### Questions:
Why don't we use config.WhenIWork in tasks?

### DIFF
--- a/tasks/timeOffRequests.js
+++ b/tasks/timeOffRequests.js
@@ -1,0 +1,124 @@
+var WhenIWork = require('wheniwork-unofficial');
+var api = new WhenIWork(KEYS.wheniwork.api_key, KEYS.wheniwork.username, KEYS.wheniwork.password);
+var moment = require('moment');
+var date_format = 'YYYY-MM-DD HH:mm:ss';
+
+var fs = require('fs');
+
+/* 
+  timeOffRequests: Returns a csv (written to file) of timeOffRequests for designated timeframe. 
+
+  Parameters
+  ----------
+  endTimeDate : String
+    Designates the latest date of the request
+    Defaults to todays date and time
+  daysBack : Number
+    How many days before the endTimeDate to request
+*/
+
+module.exports.timeOffRequests = function (endTimeDate, daysBack) {
+  var reqParams = createDateParams(endTimeDate, daysBack);
+
+  getRequestsPromise(reqParams)
+  .then(mergeResponses)
+  .then(reformatForCSV)
+  .then(exportCSV)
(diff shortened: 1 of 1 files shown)


tasks/timeOffRequests.js
```diff
@@ -0,0 +1,124 @@
...
+  getRequestsPromise(reqParams)
+  .then(mergeResponses)
+  .then(reformatForCSV)
+  .then(exportCSV)
+  .catch(function(err){
+    CONSOLE_WITH_TIME(err);
+  });
+
+  function createDatesParams(endTimeDate, totalDays) {
+    var end = endTimeDate ? moment(endTimeDate, date_format).format(date_format) : moment().format(date_format);
```
jrauschenberg: Wouldn't it be cleaner to set "end" above where you set totalDays? Then you can confidently pass both into createDatesParams.

tasks/timeOffRequests.js
```diff
@@ -0,0 +1,124 @@
+var WhenIWork = require('wheniwork-unofficial');
+var api = new WhenIWork(KEYS.wheniwork.api_key, KEYS.wheniwork.username, KEYS.wheniwork.password);
```
jrauschenberg: Should we use the config WiW here so you can test this if needed?

jrauschenberg:
I tested manually on my local computer and it worked like a charm. Other than those couple comments, I think it's good to go! 👍

tasks/timeOffRequests.js
```diff
@@ -0,0 +1,124 @@
...
+var moment = require('moment');
+var date_format = 'YYYY-MM-DD HH:mm:ss';
+
+var fs = require('fs');
+
+/* 
+  timeOffRequests: Returns a csv (written to file) of timeOffRequests for designated timeframe.
```
tongxiang: 💯  for good documentation!

tongxiang:
👍 good to merge @apackin.

tongxiang:
(How to run with console function: `node console timeOffRequests timeOffRequests`)

apackin:
@tongxiang should I change the naming so that timeOffRequests doesn't have to be typed twice somehow?

PR #139, opened 2016-06-28
handled the supervisor edgecase

mbogor:
#### What's this PR do?
Stops the grading script is the user is a supervisor/trainer.
jobs/scheduling/AttendanceSync.js
INT-194


mbogor:
hold on merging; the sample data file is undergoing major changes, after those are merged in, this file needs to be refactored again.

jobs/scheduling/AttendanceSync.js
```diff
@@ -156,6 +156,10 @@ function queryForCanvasCoursesAndAssignments(userID){
     if (!enrollmentObj.length) { 
       CONSOLE_WITH_TIME('This user has no enrollments');
       return;}
+    if(eobj[0].type=='TeacherEnrollment' || 'DesignerEnrollment'){
+      CONSOLE_WITH_TIME('THE USER ' + eobj[0].user.name + ' APPEARS TO BE A TEACHER, ABORTING GRADING');
```
jrauschenberg: You use "eobj" here but "enrollmentObj" above. Which is it? Or are they really two separate variables?

jobs/scheduling/AttendanceSync.js
```diff
@@ -156,6 +156,10 @@ function queryForCanvasCoursesAndAssignments(userID){
     if (!enrollmentObj.length) { 
       CONSOLE_WITH_TIME('This user has no enrollments');
       return;}
+    if(eobj[0].type=='TeacherEnrollment' || 'DesignerEnrollment'){
```
jrauschenberg: I think you need to put an expression on both sides of the or, as in eobj[0].type=='TeacherEnrollment' || eobj[0].type=='DesignerEnrollment'

jrauschenberg:
A couple comments. Other than that, good to merge. 👍

jobs/scheduling/AttendanceSync.js
mbogor: i fixed this in my 'fixed merge conflicts' commit

jobs/scheduling/AttendanceSync.js
mbogor: nah it works both ways, just tested it https://repl.it/C6s5

jobs/scheduling/AttendanceSync.js
mbogor: good looking out, fixed

jobs/scheduling/AttendanceSync.js
jrauschenberg: No, it doesn't seem to:

function testthis(vari){
if(vari==1 || 2){
console.log('started from the bottom now we here')
}
}

testthis(4) ==> still console.logs! check it out

mbogor: u right

tongxiang:
👍

PR #142, opened 2016-06-28
Modify retrieve and render shifts to delete and update email to support@

apackin:
https://admin.crisistextline.org/jira/secure/RapidBoard.jspa?rapidView=5&view=detail&selectedIssue=INT-184


tongxiang:
👍

PR #177, opened 2016-07-27
added catchblocks to attendancesync script

mbogor:
#### What's this PR do?
adds catch blocks to the attendancesync script
jobs/scheduling/AttendanceSync.js
run attendanceSync (change the mandrill email address please so that heather doesn't get bombarded by our testing) and attendanceSync tests
INT-305

### DIFF
--- a/jobs/scheduling/AttendanceSync.js
+++ b/jobs/scheduling/AttendanceSync.js
@@ -1,3 +1,4 @@
+'use strict';
 var CronJob = require('cron').CronJob;
 var canvas = require('../../canvas.js');
 var bignumJSON = require('json-bignum');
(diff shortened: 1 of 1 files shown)


jobs/scheduling/AttendanceSync.js
```diff
@@ -168,13 +175,22 @@ function queryForCanvasCoursesAndAssignments(userID){
     .then(function(assignment){
```
apackin: If you return canvas.scrapeCanvasA(courseID, assignmentQuery)
you could continue with the then below and only have one catch block.

apackin:
👍

tongxiang:
👍

jobs/scheduling/AttendanceSync.js
mbogor: i want to preserve the two different catch blocks because:
- they have two different messages which makes it easier for debugging
- if i .then off of scrapeCanvasA and the call fails, i lose courseID which is useful

PR #178, opened 2016-08-05
add routes for MIA 2 weeks and filtered

apackin:
#### What's this PR do?
Creates routes for the Baby CRM
https://admin.crisistextline.org/jira/secure/RapidBoard.jspa?rapidView=5&view=detail&selectedIssue=INT-313

### DIFF
--- a/api_wiw/WiWRequests.js
+++ b/api_wiw/WiWRequests.js
@@ -1,0 +1,62 @@
+'use strict';
+
+const api = CONFIG.WhenIWork;
+
+function getTimeoff (params) {
+  return getTimeoffRequests(params)
+  .then(mergeResponses)
+  .catch(function(err){
+    CONSOLE_WITH_TIME(err);
+  });
+}
+
+function getTimeoffRequests (params) {
+  const allResponses = [];
+  let lastReqId;
+  return new Promise(function(resolve, reject){
+    // Scoped a recursive IIFE into the promise so that I can resolve out.
+    (function makeGetRequest () {
+      api.get('requests', params, function(response) {
+        if (!response.requests) reject(response);
+        allResponses.push(response);
+        // The WiW API responds with a maximum of 200 requests, so if we received fewer (conservatively 100) we can resolve.
+        if (response.requests.length < 100) resolve(allResponses);
+        // If we received over 100 requests make a new request starting from the last object Id we received. 
+        else {
+          lastReqId = response.requests[response.requests.length-10].id;
+          params.max_id = lastReqId;
+          makeGetRequest ();
+        }
+      });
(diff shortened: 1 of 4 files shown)


api_wiw/WiWRequests.js
```diff
@@ -0,0 +1,62 @@
...
+      });
+    })();
+  });
+}
+
+
+function mergeResponses (responses){
+  const allUsers = {};
+  const allRequests = {};
+  // merge the users and time off requests from all responses
```
jrauschenberg: Do allUsers and allRequests need to be consts? It seems like they're just returned out of the function so don't need to be.

www/data/index.js
```diff
@@ -16,4 +19,107 @@ router.get('/wiw/timeoff', function (req, res) {
...
+      console.log('err', err);
+      res.status(500).send(err);
+    });
+  }
+  else res.status(403).send('Access denied.');
+});
+
+function convertJsonToCSV(JSONarray) {
+  let CSVFormattedString = '';
+  let values;
```
jrauschenberg: could you just say let values = []  up here?

www/data/index.js
```diff
@@ -16,4 +19,107 @@ router.get('/wiw/timeoff', function (req, res) {
...
+router.get('/wiw/Mia2WeeksLastLogin', function (req, res) {
+  const options = {
+    uri: 'https://app.crisistextline.org/ajax/stats/absentees',
+    qs: {
+        token: KEYS.platform_secret_key
+    },
+    headers: {
+        'User-Agent': 'Request-Promise'
+    },
+    json: true // Automatically parses the JSON string in the response
```
jrauschenberg: I love json: true it saves me so many times

www/data/index.js
```diff
@@ -16,4 +19,107 @@ router.get('/wiw/timeoff', function (req, res) {
...
+  };
+
+  if (KEYS.dataRoute && req.query.token === KEYS.dataRoute) {
+
```
jrauschenberg: extra space?

jrauschenberg:
Minor changes, otherwise good to go! 👍

www/data/index.js
tongxiang: Great to know. We should add this throughout the codebase.

tongxiang:
@assaf--can you add "KEYS.dataRoute" into `keys.default.js`? Thanks!

www/data/index.js
apackin: I could do that, but I need to reset it in the for loop for every key anyway so it wouldn't do much.

tongxiang:
👍 good to merge!

PR #179, opened 2016-08-11
Churn users

mbogor:
#### What's this PR do?
refactored remove churned users script to include less api calls
try login in as a newUser, old user, and old deleted user
line 102 in schedule/index.js i use the request module instead of wheniwork unofficial because for some reason the wheniwork didn't return deleted users when i used it. if you have any ideas about that and/or a better way of doing this, i'd love to hear about it
INT-226, INT-308


www/scheduling/index.js
```diff
@@ -83,51 +84,107 @@ function getTimezones(req, res) {
...
+  });
+}
+
+// retrieves all WiW users and checks for a particular user
+function allUsers(email, altEmail, callback){
+  return new Promise (function(resolve, reject){
+
+    var options = {
+        method: 'get',
+        url: 'https://api.wheniwork.com/2/users?include_objects=false&show_deleted=true',
```
apackin: You know what it is

test/loginUserCreation.js
```diff
@@ -7,17 +11,17 @@ describe('Login and user creation', function () {
...
         assert.equal(user.email, 'amudantest@test.com');
         done();
       });
     });
-
-    it('Creates a new user if one doesn\'t exist', function(done) {
-      this.timeout(4000);
-      checkUser('test2@test.com', 'Testy2', 'McTesterson2', function(user) {
-        assert.equal(user, 'test2@test.com');
-        done();
```
apackin: Do we want to write a working test for this?

test/loginUserCreation.js
```diff
@@ -1,4 +1,8 @@
 'use strict';
+
+global.KEYS = require('../keys.js');
+global.CONFIG = require('../config.js')
```
apackin: Should be covered by test/index

test/loginUserCreation.js
```diff
@@ -7,17 +11,17 @@ describe('Login and user creation', function () {
     it('Doesn\'t create a user that already exists', function(done) {
       this.timeout(4000);
       checkUser('amudantest@test.com', 'Amudan', 'Test', function(user) {
+        console.log('user email ',user.email)
```
apackin: remove?

tasks/removeChurned.js
```diff
@@ -11,101 +11,35 @@ const churnedUserEmailList = require('../churnedUserEmailList');
...
 function getUsersToClean() {
   return new Promise(function (resolve, reject) {
     var users, email;
     var uidsToClean = [];
     api.get('users?include_objects=false', function (res) {
+      if (!res) reject('Call to get WiW users failed');
       users = res.users;
 
-      users.forEach(function (e, i, arr) {
+      users.forEach(function (each, item, arr) {
```
apackin: <3

test/index.js
```diff
@@ -10,7 +10,7 @@ global.CONFIG = require('../config.js');
 var recurOpenShifts = require('./recurOpenShifts.js');
 var colorize = require('./colorize.js');
 var attendanceSync = require('./attendanceSync.js');
-var loginUserCreation = require('./loginUserCreation.js');
+// var loginUserCreation = require('./loginUserCreation.js');
```
apackin: This whole test file isn't working? Can we fix it?

www/scheduling/index.js
```diff
@@ -83,51 +84,101 @@ function getTimezones(req, res) {
...
+    });
+  });
+}
+
+// retrieves all WiW users and checks for a particular user
+function allUsers(email, altEmail, callback){
+  return new Promise (function(resolve, reject){
+    api.get('users', {include_objects: false, show_deleted: true}, function(data){
+        if (!data) {
+            reject('No data returned from WiW');
```
apackin: I'm not that familiar with include_object: false -> When there is a bad response does it not return and object with a message param that would pass this if?

www/scheduling/index.js
```diff
@@ -83,51 +84,101 @@ function getTimezones(req, res) {
...
+      if (res) resolve(res);
+      else reject('User reactivation failed');
+    });
+  });
+}
+
+// retrieves all WiW users and checks for a particular user
+function allUsers(email, altEmail, callback){
+  return new Promise (function(resolve, reject){
+    api.get('users', {include_objects: false, show_deleted: true}, function(data){
```
apackin: I like 'response' over 'data' /shrug

www/scheduling/index.js
```diff
@@ -83,51 +84,101 @@ function getTimezones(req, res) {
...
+        if (!data) {
+            reject('No data returned from WiW');
+        } else {
+          var users = data.users;
+          for (var i in users) {
+            // existing users
+            if (users[i].email === email && !users[i].is_deleted || users[i].email === altEmail && !users[i].is_deleted) {
```
apackin: You can DRY this by adding 
if (!users[i].is_deleted) as a sub-case of the email checks

apackin: if (users[i].email === email || users[i].email === altEmail) {
   if (!users[i].is_deleted) resolve....
  else reactivate...

www/scheduling/index.js
```diff
@@ -83,51 +84,101 @@ function getTimezones(req, res) {
...
+                  callback(users[i]);
+                  resolve(users[i]);
+                })
+                .catch(function(error){
+                  CONSOLE_WITH_TIME('User', users[i].login_id,'reactivation failed', error)
+                });
+            } 
+          }
+          // we didn't find the user in WiW records, we need to create a newUser
+          resolve();
```
apackin: else here?

www/scheduling/index.js
```diff
@@ -83,51 +84,101 @@ function getTimezones(req, res) {
...
+    });
+  });
+}
+
+// creates a new user
+function createUser(newUser, callback){
+  return new Promise(function(resolve, reject){
     stathat.increment('Scheduling - Accounts Created', 1);
-    /**
-      At this point, we didn't find the user so let's create it.
```
apackin: Put these notes in the promise?

jobs/scheduling/dropMeltedUsers.js
```diff
@@ -1,3 +1,4 @@
+"use strict";
```
apackin: ' instead of " :-D

tasks/removeChurned.js
```diff
@@ -11,101 +11,35 @@ const churnedUserEmailList = require('../churnedUserEmailList');
...
         try {
-          email = JSON.parse(e.notes).canonicalEmail;
```
apackin: I prefer doing this with a regex search before parsing like:
email = /canonicalEmail/.test(user.notes) ? 
        JSON.parse(each.notes).canonicalEmail : 
        each.email;

tasks/removeChurned.js
apackin: can we call this (user, idx, arr) ? 
I'm not sure what item means -> is that a CS term?

test/loginUserCreation.js
```diff
@@ -7,17 +11,17 @@ describe('Login and user creation', function () {
...
-
-    it('Creates a new user if one doesn\'t exist', function(done) {
-      this.timeout(4000);
-      checkUser('test2@test.com', 'Testy2', 'McTesterson2', function(user) {
-        assert.equal(user, 'test2@test.com');
-        done();
-      });
-    });
-
+    //this test can't be run more than once; we can only create a new user once, afterwards it's an existing user. if we delete that user, it will just be an existing deleted user and the new version of the login script will just reactivate that user
```
apackin: When testing this should be using the fake version of WiW so the user should never actually be created?

www/scheduling/index.js
```diff
@@ -83,51 +84,101 @@ function getTimezones(req, res) {
...
-    for (var i in users) {
-      if (users[i].email == email || users[i].email == altEmail) {
-        callback(users[i]);
-        return;
-      }
-    }
+// reactivating a deleted user
+function reactivate(user){
+  return new Promise(function(resolve, reject){
+    api.update('users/'+user.id, {reactivate:true}, function(res){
```
apackin: `users/${user.id}`

jobs/scheduling/dropMeltedUsers.js
mbogor: ? what do you mean

test/loginUserCreation.js
mbogor: the problem is that this test can't be run more than once; we can only create a new user once, afterwards it's an existing user. if we delete that user, it will just be an existing deleted user and the new version of the login script will just reactivate that user instead of creating a new one. i think i'm going to rewrite this to check that a deleted user gets reactivated. what do you think?

test/loginUserCreation.js
mbogor: yas my b, forgot to take this out

test/loginUserCreation.js
mbogor: yas

tasks/removeChurned.js
mbogor: oh no, i think in traditional cs people use i and j for variables because most math theorems use i and j as stand-in integers. index makes more sense, i'll change it

test/index.js
mbogor: when i run this file individually the tests pass, but when i run it with the suite, for some reason this file runs twice and i get the "done called multiple times error"

www/scheduling/index.js
mbogor: good point, will check it out

www/scheduling/index.js
mbogor: good point, i do think we mostly use response instead of data throughout elmer, will conform

www/scheduling/index.js
mbogor: no, because there are several cases possible: deleted user w matching email, deleted user w/o the matching email, existing user with the matching email, existing user w/o the matching email, not matching deleted user, not matching existing user and completely new/non-matching user.

www/scheduling/index.js
mbogor: yas, good call

jobs/scheduling/dropMeltedUsers.js
mbogor: got it

tasks/removeChurned.js
mbogor: why? is it better or is it just a preference?

test/loginUserCreation.js
mbogor: true, i'll double check that it's actually doing that

www/scheduling/index.js
mbogor: yas, will change

tasks/removeChurned.js
apackin: I don't have a great reason. I've read that you should avoid try-catch in loops, but don't remember the reason. 
+ It makes me uneasy to write code that is supposed to error when performing correctly with the intention of catching those errors. The try catch way is probably actually more performant because we almost always have canonicalEmail's in the notes ¯\_(ツ)_/¯

www/scheduling/index.js
tongxiang: @mbogor flagging this as unchanged

tongxiang:
👍 good to merge!

PR #180, opened 2016-08-17
added platformCreation script notice and fixed get parameters in time…

mbogor:
#### What's this PR do?
- adds a message alerting the graduation slack channel when the grad script is running
- fixed whenIwork.get parameters call in timeOff requests 
- calls only for "workflow_state": "available" courses' grade logs; significantly reduces the amount of api calls necessary to make
INT - 291
#### Questions:

…off requests

### DIFF
--- a/jobs/scheduling/TimeOffRequests.js
+++ b/jobs/scheduling/TimeOffRequests.js
@@ -15,9 +15,9 @@ handleTimeOffRequests();
 
 function handleTimeOffRequests() {
   var timeOffSearchParams = createTimeOffSearchParams();
-
+  timeOffSearchParams.include_objects=false;
   //Get all time off requests within timeOffSearchParams
-  WhenIWork.get('requests?include_objects=false', timeOffSearchParams, function(response){
+  WhenIWork.get('requests', timeOffSearchParams, function(response){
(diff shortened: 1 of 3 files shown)


jobs/scheduling/TimeOffRequests.js
```diff
@@ -49,8 +49,8 @@ function filterRequestsAndHandleShifts (allRequests) {
 
 function retrieveOverlappingShifts (request) {
   var shiftSearchParams = createShiftSearchParams(request);
-
-  WhenIWork.get('shifts?include_objects=false', shiftSearchParams, function(response) {
+  shiftSearchParams.include_objects=false;
```
jrauschenberg: I wonder if you could combine this function with createTimeOffSearchParams. I know it has extra params, but I think WiW will ignore them if they don't apply and/or aren't filled in. Worth thinking about.

jrauschenberg:
One comment. Otherwise good to go. 👍

jobs/scheduling/TimeOffRequests.js
mbogor: i like that, fixed it.

tongxiang:
👍

PR #182, opened 2016-08-18
Changed all queries including 'include_objects=false' to put that in …

jrauschenberg:
#### What's this PR do?
Changes all calls to the WiW API that used to include "?include_objects=false" as part of the query string. Those were failing because the WiW unofficial API was adding a question mark of its own. Instead, all calls now include "include_objects: false" in a parameters object.
Search globally for "include_objects" to see all the places where it is used.
Npm test and/or node any file that uses an "include_objects: false" object to see that extra objects are not returned. I tested each altered file with this process:
1. Add global.KEYS and global.CONFIG definitions to the file I want to test.
2. Add a global.KEYS definition to api_wiw/WiWCCApi.js.
3. Add a console.log to the callback function that runs on the response from WiW to see what is returned. Comment out anything in the callback function that I don't want to run (I just commented out everything other than the console.log to be safe).
4. Ensure that the function that I want to run in order to make the WiW API call is going to run. To do this, sometimes I needed to call the function in the file I wanted to test.
5. Confirm that extra objects (e.g. locations for a 'shifts' get call) are not returned in the terminal.

[INT-297
](https://admin.crisistextline.org/jira/browse/INT-297)#### Questions:

…a params object instead of a query string, as the query string was turning out malformed.

### DIFF
--- a/tasks/schedule.js
+++ b/tasks/schedule.js
@@ -9,12 +9,13 @@ const out_format = 'ddd h:mm A';
 
 module.exports.dumpSchedules = function () {
   // We want to see shifts 8 days out from now.
-  var params = {
+  let params = {
     location_id: CONFIG.locationID.regular_shifts,
-    end: '+8 days'
+    end: '+8 days',
+    include_objects: false
   };
 
-  api.get('shifts?include_objects=false', params, function (data) {
+  api.get('shifts', params, function (data) {
     var user_shifts = {};
     var shift;
 
@@ -33,7 +34,10 @@ module.exports.dumpSchedules = function () {
     }
 
     // Now we need to get the email addresses
-    api.get('users?include_objects=false', function (data) {
+    let params = {
+      include_objects: false
+    };
+    api.get('users', params, function (data) {
       var user;
       for (var i in data.users) {
         user = data.users[i];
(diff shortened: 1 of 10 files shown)


tasks/schedule.js
```diff
@@ -76,10 +80,11 @@ module.exports.clearProd = function() {
       end: end_time,
       location_id: CONFIG.locationID.test,
       unpublished: true,
-      include_allopen: true
+      include_allopen: true,
+      include_objects: false
     };
 
-    api.get('shifts?include_objects=false', query, function (data) {
+    api.get('shifts, query, function (data) {
```
mbogor: close the quotes on shifts

test/helpers/base.js
```diff
@@ -13,18 +13,12 @@ WhenIWork.prototype.get =  function(term, params, cbFunction){
   //is the second argument. May need to change this
   //in the future depending on what gets passed to 'get'.
   cbFunction = [].slice.call(arguments).pop();  
-  if (term === 'users?include_objects=false') {
-    cbFunction(sampleData.usersResponse);
-  }
-  else if (term === 'users') {
+  if (term === 'users') {
```
mbogor: good thinking

mbogor:
👍

tongxiang:
👍 let's merge and deploy on Monday!

jrauschenberg:
Okay, I'm going to merge this right now, unless you have any last-minute
objections!

John

PR #185, opened 2016-08-26
Changed all urls including '/2/' at the beginning to delete the redun…

jrauschenberg:
#### What's this PR do?
There are lots of places in our code where we include '/2/shifts/', for example, as a url in our parameters. However, these urls eventually get sent to the When I Work unofficial API, which automatically begins every url thusly: 
```this.endpoint = 'https://api.wheniwork.com/2';```
so as far as I can tell the '2' is redundant, and may have even been causing some problems. I deleted the '/2' everywhere I found it.
Do a global find for '/2' or "/2" or `/2` to make sure I didn't miss anything.
Run any of the affected files in Node and console.log the results to ensure that their WiW calls are working. Run npm test as well.
[INT-356
](https://admin.crisistextline.org/jira/browse/INT-356)#### Questions:

…dant 2.

### DIFF
--- a/jobs/scheduling/MergeOpenShifts.js
+++ b/jobs/scheduling/MergeOpenShifts.js
@@ -91,7 +91,7 @@ function makeBatchPayloadRequestsToMergeOpenShifts(arrayOfShiftsForSameTimeInt, 
 
         var shiftUpdateRequest = {
           'method': 'PUT',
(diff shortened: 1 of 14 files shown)


mbogor:
👍

tongxiang:
👍

jrauschenberg:
I'm not going to merge this yet, because it needs the WiW unofficial update (PR 7 in that repo) to happen first.

PR #187, opened 2016-08-29
Intergrate google sheet api and build 2 week email functionality

apackin:
#### What's this PR do?
Make the data routes more descriptive 
Modularize the helper functions
Enable google sheets integration
www/data/index.js and follow the functions calls out of there
You should be able to hit all of the included routes when running www/app.js
Be careful with /email/MIA2WeeksWithoutTimeoff because it actually sends emails. I disabled the mass emailing functionality until we decide to move forward with that. To send individual emails include a body with the post like
{
  "users": [
      {"first_name": "A", "last_name": "P", "email": "email@gmail.com"},
      {"first_name": "A", "last_name": "P", "email": "email@gmail.com"}
    ]
}

You can update the sheet No Contact List in https://docs.google.com/spreadsheets/d/1j49XjT1xcoN8Q5bn-EFA5vBGCgXjgUc3j8CcDQfzTW8/edit#gid=1722145441
and those email addresses should not get emailed. 

https://admin.crisistextline.org/jira/secure/RapidBoard.jspa?rapidView=5&view=detail&selectedIssue=INT-353

### DIFF
--- a/www/data/googleSheets.js
+++ b/www/data/googleSheets.js
@@ -1,0 +1,112 @@
+'use strict';
+
+const readline = require('readline');
+const google = require('googleapis');
+const GoogleAuth = require('google-auth-library');
+// If modifying these scopes, delete your previously saved credentials
+// at ~/.credentials/sheets.googleapis.com-nodejs-quickstart.json
+const SCOPES = ['https://www.googleapis.com/auth/spreadsheets.readonly'];
+
+/*
+ * @param {String} sheetId The id of the G sheet (last part of URL).
+ * @param {String} sheetName The name of the specific sheet to access.
+ * @param {String} cellRange The range of cells to retrieve from that sheet.
+*/ 
+function getGsheetRange(sheetId, sheetName, cellRange) {
+  const range = `${sheetName}!${cellRange}`;
+  const specificRequest = createRequestPromise(sheetId, range);
+// Authorize a client with the loaded credentials, then call the Google Sheets API.
+  return authorize(KEYS.gSheets, specificRequest);
+}
+
+/**
+ * Create an OAuth2 client with the given credentials, and return a Promise
+ * for the response.
+ *
+ * @param {Object} credentials The authorization client credentials.
+ * @param {Promise} promise The promise to create with the authorized client.
+ */
+function authorize(credentials, promise) {
+  const clientSecret = credentials.installed.client_secret;
+  const clientId = credentials.installed.client_id;
+  const redirectUrl = credentials.installed.redirect_uris[0];
+  const auth = new GoogleAuth();
+  const oauth2Client = new auth.OAuth2(clientId, clientSecret, redirectUrl);
+
+  // Check if we have previously stored a token.
+  if (KEYS.gSheets.token) {
+    oauth2Client.credentials = KEYS.gSheets.token;
+    return promise(oauth2Client);
+  } else {
+    return getNewToken(oauth2Client, promise);
+  }
+}
+
+/**
+ * Create a request for the specific sheet and cell range
+ */
+function createRequestPromise (spreadsheetId, range) {
+  return function (auth) {
+    return new Promise(function(resolve, reject) {
+      const sheets = google.sheets('v4');
+      sheets.spreadsheets.values.get({
+        auth: auth,
+        spreadsheetId,
+        range,
+      }, function(err, response) {
+        if (err) {
+          reject('The API returned an error: ' + err);
+        }
+        if (response.values.length === 0) {
+          reject('No data found.');
+        } else {
+          resolve(response);
+        }
+      });
+    });
+  };
+}
+
+/**
+ * Get and store new token after prompting for user authorization, and then
+ * return a Promise for the response.
+ *
+ * @param {google.auth.OAuth2} oauth2Client The OAuth2 client to get token for.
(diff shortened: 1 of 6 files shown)


www/data/googleSheets.js
```diff
@@ -0,0 +1,113 @@
...
+      const sheets = google.sheets('v4');
+      sheets.spreadsheets.values.get({
+        auth: auth,
+        spreadsheetId,
+        range,
+      }, function(err, response) {
+        if (err) {
+          reject('The API returned an error: ' + err);
+        }
+        const rows = response.values;
```
jrauschenberg: Do you need 'rows' here? Couldn't you just check response.values.length?

www/data/googleSheets.js
```diff
@@ -0,0 +1,113 @@
...
+ * This should store token to disk be used in later program executions.
+ *
+ * @param {Object} token The token to store to disk.
+ */
+function storeToken(token) {
+  console.log('Token for use ' + JSON.stringify(token));
+  KEYS.gSheets.token = token;
+}
+
+module.exports = {getGsheetRange};
```
jrauschenberg: Any specific reason you're making the getGsheetRange function part of an object here?

www/data/helpers.js
```diff
@@ -0,0 +1,127 @@
...
+function doNotContact () {
+  return gSheets.getGsheetRange(KEYS.gSheets.sheetIds.babyCRM, 'No Contact List', 'A2:A')
+  .then(res => {
+    const doNotContactObject = {};
+    res.values.forEach(row => {
+      doNotContactObject[row[0]] = 'NoContact';
+    });
+    return doNotContactObject;
+  })
+  .catch(err => console.log('err', err));
```
jrauschenberg: Maybe a CONSOLE_WITH_TIME?

www/data/index.js
```diff
@@ -31,94 +40,52 @@ router.get('/wiw/Mia2WeeksLastLogin', function (req, res) {
...
-    .then(CSVString => {
-      res.send(CSVString);
-    })
-    .catch(err => {
-      res.status(500).send(err);
-    });
-  }
-  else res.status(403).send('Access denied.');
+  request(options)
+  .then(helpers.convertJsonToCSV)
```
jrauschenberg: Do you want to be calling this function here? What returns if you don't call it?

www/data/index.js
```diff
@@ -31,94 +40,52 @@ router.get('/wiw/Mia2WeeksLastLogin', function (req, res) {
...
+  .then(CSVString => {
+    res.send(CSVString);
+  })
+  .catch(err => {
+    res.status(500).send(err);
```
jrauschenberg: Is there a way to have a more descriptive error here?

www/data/index.js
```diff
@@ -31,94 +40,52 @@ router.get('/wiw/Mia2WeeksLastLogin', function (req, res) {
...
-          JSON.parse(user.notes).canonicalEmail : 
-          user.email;
-        usersWhoRequestTimeOff[email] = 1;
-      }
-      const usersWithNoExcuse = responses[1].filter(user => {
-        return !usersWhoRequestTimeOff[user.email];
-      });
+router.get('/MIA2WeeksWithoutTimeoff', function (req, res) {
+  helpers.getUsersFilterRequestedTimeOff()
+  .then(helpers.convertJsonToCSV)
```
jrauschenberg: Same Qs as above, lines 43-50.

www/data/googleSheets.js
apackin: I'm thinking we will want to export more functionality in the future. It's simpler to do that if we start by exporting an object instead of the function itself.

www/data/index.js
```diff
@@ -31,94 +40,52 @@ router.get('/wiw/Mia2WeeksLastLogin', function (req, res) {
...
-      if (elem.hasOwnProperty(k)) { 
-        value = `${elem[k]}`;
-        values.push(value.replace(/,/g, ''));
-      }
-    }
-    CSVFormattedString += `${values.join(',')}\n`;
-  }
-  return CSVFormattedString;
+function send2WeeksMissedEmail (usersToEmail) {
+  return usersToEmail.map(helpers.compose2WeekEmailAndSend);
```
jrauschenberg: Nice use of map!

www/data/index.js
apackin: .then runs the function passed to it as the first parameter. The return of that function is passed down the chain.

jrauschenberg:
👍

www/data/googleSheets.js
tongxiang: How will this store the token to disk? Won't this just keep it in memory, attached to the `KEYS` object already read in by Node?

www/data/helpers.js
```diff
@@ -0,0 +1,134 @@
...
+}
+
+function compose2WeekEmailAndSend (user) {
+  const message = {
+    subject: 'We miss you!',
+    html: composeEmail(user) + user.email,
+    from_email: 'support@crisistextline.org',
+    from_name: 'Crisis Text Line',
+    to: [{
+        email: 'apackin@gmail.com',
```
tongxiang: guessing we'll change this when we have finalized copy

tongxiang:
👍

PR #188, opened 2016-09-01
Updated WiW unofficial

apackin:
We removed the '/2's from elmer without updating the wiw version yet.


tongxiang:
👍

jrauschenberg:
Hmm...the update was merged into the WiW repo, but yes, the version was still 0.2.3 at that point. It should be 0.2.4 now, right?

PR #189, opened 2016-09-01
error handling to remove churned script, new job to handle pending users

mbogor:
#### What's this PR do?
* added error handling to remove churned users script
* added a script that reviews pending WiW users and if it finds WiW deleted user w the same email it reactivates the deleted user and deletes the pending user object; if it finds an already active WiW user it deleted the pending user object

* removePending.js

* add keys and config to the top of the removePending file
* comment out the cronjob
* node removePending.js in the console
* before running the job make sure that there are WiW pending users to be dealt with

https://admin.crisistextline.org/jira/browse/INT-349


jobs/scheduling/removePending.js
```diff
@@ -0,0 +1,75 @@
...
+	Some churned users try to log bak into WiW directly rather than CTL's single sign on. That bypasses all of our proper reactivation scripts and creates a new pending user object. We, however, prefer to reactivate deleted users rather than create redundant new profiles.
+**/
+const api = CONFIG.WhenIWork;
+const CronJob = require('cron').CronJob;
+
+new CronJob(CONFIG.time_interval.pending_users, go, null, true);
+go();
+// get this party started
+function go(){
+	getPendingUsers()
```
apackin: Beautiful modular promise chain

jobs/scheduling/removePending.js
```diff
@@ -0,0 +1,75 @@
...
+		CONSOLE_WITH_TIME('WiW Pending Users Reativation Failed, Error', res);
+	});
+}
+
+// get WiW approval pending users
+function getPendingUsers(){
+	return new Promise(function(resolve, reject){
+		api.get('users', function(res){
+			if (res.message) reject(`Call to get WiW pending users failed: ${res.message}`);
+			var pending = {};
```
apackin: const if you want

jobs/scheduling/removePending.js
```diff
@@ -0,0 +1,75 @@
+'use strict';
+/**
+	Some churned users try to log bak into WiW directly rather than CTL's single sign on. That bypasses all of our proper reactivation scripts and creates a new pending user object. We, however, prefer to reactivate deleted users rather than create redundant new profiles.
```
apackin: It looks like your tab setting changed back to more than 2 spaces per tab

jobs/scheduling/removePending.js
```diff
@@ -0,0 +1,75 @@
...
+	});
+}
+
+// get WiW approval pending users
+function getPendingUsers(){
+	return new Promise(function(resolve, reject){
+		api.get('users', function(res){
+			if (res.message) reject(`Call to get WiW pending users failed: ${res.message}`);
+			var pending = {};
+			res.users.forEach(function(user){
```
apackin: It might make sense to separate the forEaches logic from the api call promise so that we can test them independently if we want to later.

removePending.js
```diff
@@ -0,0 +1,75 @@
+'use strict';
```
apackin: Is this the same file again?

jobs/scheduling/removePending.js
```diff
@@ -0,0 +1,75 @@
...
+				} 
+			});
+			resolve(reactivate);
+		});
+	});
+}
+
+// reactivating a deleted user rather than approving a reduntant pending user
+function reactivate(list){
+	console.log('lenght', list.length, 'list', list);
```
apackin: lenght

jobs/scheduling/removePending.js
```diff
@@ -0,0 +1,75 @@
...
+		});
+	});
+}
+
+function getDeletedUsers(pending){
+	return new Promise(function(resolve, reject){
+		api.get('users', {show_deleted: true}, function(res){
+			if (res.message) reject(`Call to get WiW deleted users failed: ${res.message}`);
+			var reactivate = [];
+			// if a pending user matches a user we have previously deleted, reactivate the deleted user and reject a pending user
```
apackin: This comment refers to line 46? Can we move it down?

jobs/scheduling/removePending.js
```diff
@@ -0,0 +1,75 @@
...
+			res.users.forEach(function(user){
+				if (!user.is_deleted && pending[user.email]) {
+					// reject the pending user
```
apackin: reject the pending user because they already exist
Should we be looking at the canonical emails for this? I thought we add an admin+ to their 'official' WiW email address.

apackin:
Looks good. Mainly style comments edits. 
The only thing I'm wondering about for the functional logic is that we're storing fake emails for many of the WiW users as the main emails on their account by 'admin+${email}@crisistextline.org' and adding their original email as the canonical email in their notes. So you may need to check against that in the forEach of getDeletedUsers.

removePending.js
mbogor: yep, pretty bizarre, removed, thanks

jobs/scheduling/removePending.js
mbogor: yep, thanks good catch

jobs/scheduling/removePending.js
mbogor: yep, makes sense to me; moved

jobs/scheduling/removePending.js
mbogor: yes, done.

jobs/scheduling/removePending.js
mbogor: no, it's still 2. maybe this is some github weirdness

jobs/scheduling/removePending.js
mbogor: we create admin accounts when the user already has a regular non admin+ account, so this script will just reactivate their original accounts rather than their admin accounts

jobs/scheduling/removePending.js
tongxiang: *back

Also, would be good to explicitly say what this job does in the comments

jobs/scheduling/removePending.js
```diff
@@ -0,0 +1,74 @@
+'use strict';
+/**
+	Some churned users try to log bak into WiW directly rather than CTL's single sign on. That bypasses all of our proper reactivation scripts and creates a new pending user object. We, however, prefer to reactivate deleted users rather than create redundant new profiles.
+**/
+const api = CONFIG.WhenIWork;
+const CronJob = require('cron').CronJob;
+
+// new CronJob(CONFIG.time_interval.pending_users, go, null, true);
```
tongxiang: We want to uncomment this before committing, right? 

And do we want `go()` to run on start?

jobs/scheduling/removePending.js
```diff
@@ -0,0 +1,74 @@
...
+					rejectUser(pending[user.email]);
+				} 
+			});
+			// if a pending user matches a user we have previously deleted, reactivate the deleted user and reject a pending user
+			resolve(reactivate);
+		});
+	});
+}
+
+// reactivating a deleted user rather than approving a reduntant pending user
```
tongxiang: minor :) 

"redundant"

jobs/scheduling/removePending.js
```diff
@@ -0,0 +1,74 @@
...
+function reactivate(list){
+  return new Promise(function(resolve, reject){
+  	list.forEach(function(user){
+  		CONSOLE_WITH_TIME('Reactivating user', Object.keys(user)[0]);
+	    api.update(`users/${user.id}`, {reactivate:true}, function(res){
+	      if (res.message) reject(`User reactivation failed: ${res.message}`);
```
tongxiang: Does `res.message` include `user.id`? If not, can we include it in the error log?

tongxiang:
Well-written, elegant, and easy-to-understand code!

Minor comments. Also, if you could add notes on how you tested this, that'd be helpful.

jobs/scheduling/removePending.js
mbogor: uncommented. and i call go() under the cron job

mbogor:
@tongxiang done, thanks! is it good to merge now?

PR #190, opened 2016-09-02
Notify CCs and delete open shifts after WiW confirms recurring shifts

apackin:
#### What's this PR do?
We were notifying CC's that their shifts have been scheduled before they were successfully recurred. So if recurring failed they would get notified every time we tried to recur their shifts. This has been modified so we only notify them (and 
1. Add this to the top of the file:
global.KEYS = require('../../keys.js');
global.CONFIG = require('../../config.js');
CONFIG.locationID.regular_shifts = CONFIG.locationID.test;
console.log('testing');

2. Create a test shift in wheniwork

3. node RecurShifts.js and you should get an email. 

https://admin.crisistextline.org/jira/secure/RapidBoard.jspa?rapidView=5&view=detail&selectedIssue=INT-362
#### Other:
There's separate small formatting bug that is logging errors, but I don't believe is causing other issues. I added it to the backlog.

### DIFF
--- a/jobs/scheduling/RecurShifts.js
+++ b/jobs/scheduling/RecurShifts.js
@@ -33,6 +33,7 @@ function recurNewlyCreatedShifts() {
                   };
 
   WhenIWork.get('shifts', postData, function(response) {
+    const usersWithNewShifts = response.users;
     var allShifts = response.shifts;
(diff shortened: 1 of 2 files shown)


jrauschenberg:
👍

PR #192, opened 2016-09-06
Modify sort to use question numbers

apackin:
#### What's this PR do?
Change Sherlock sorting to be based on the questions title string instead of id #.
https://admin.crisistextline.org/jira/secure/RapidBoard.jspa?rapidView=5&view=detail&selectedIssue=INT-365


tongxiang:
👍

PR #193, opened 2016-09-08
Filter email addresses regardless of case

apackin:
#### What's this PR do?
Unified letter casing between email addresses stored in different locations
https://admin.crisistextline.org/jira/secure/RapidBoard.jspa?rapidView=5&view=detail&selectedIssue=INT-382


jrauschenberg:
👍

PR #198, opened 2016-09-22
Jenkins churned

mbogor:
#### What's this PR do?
creates a jenkins route to automate the churned users jobs
www/jenkins and task/removechurned
curl post strings to jenkins/churned route (don't forget to include token in the query)
i included a sample curling request in the JIRA subtask 
the data you pass should be in the "number email number email" format since the real data will include a data of when the user was churned followed by their email.
comment out the actual deleting function in task/removechurned since you don't want to actually send delete requests to WhenIWork lines 45-48
INT-320 INT-352


tasks/removeChurned.js
```diff
@@ -1,25 +1,31 @@
...
+const express = require('express');
+const router = express.Router();
 
-  Then run: node console removeChurned go.
-**/
-module.exports.go = function () {
-  getUsersToClean()
+
+router.post('/', function(req, res){
+  var emails = req.body.split(' ').filter((item, n)=>{if (n%2!==0) return item});
```
jrauschenberg: "n" isn't a very clear variable name.

tasks/removeChurned.js
```diff
@@ -35,12 +41,14 @@ function getUsersToClean() {
...
       });
       // delete the user and all user's shifts 
-      uidsToClean.forEach(function(id){
+      uidsToClean.forEach(function(id, index){
         api.delete(`users/${id}?delete_shifts=true`, function(res){
           if(res.error) CONSOLE_WITH_TIME(`CHURNED DELETION OF USER ${id} FAILED:`, res.error)
-          CONSOLE_WITH_TIME('Deleted user', id)
+          else CONSOLE_WITH_TIME('Deleted user', id)
         });
+        if(index==uidsToClean.length-1) resolve();
```
jrauschenberg: Why the double equals?

www/jenkins/index.js
```diff
@@ -0,0 +1,24 @@
+'use strict';
+const express   = require('express');
+const router = express.Router();
+const app = express();
+const bodyParser = require('body-parser');
```
jrauschenberg: Are you using req.body in this file?

mbogor: yes, like 8

tasks/removeChurned.js
mbogor: fixed, thanks!

tasks/removeChurned.js
mbogor: changed!

tongxiang:
Two thoughts: 

1) if we're turning `removeChurned` into a route, we should take it out of the `tasks` directory
2) is there utility to keeping it as a task, as well? That way, we could run it without hitting it as a route?

mbogor:
@tong i moved the file and added email notifications; is it good to merge?

tongxiang:
👍

PR #191, opened 2016-09-06
Modified dropMeltedUsers and updateCanvas to distinguish between firs…

jrauschenberg:
#### What's this PR do?
Creates a drop melted users process that begins when a trainer drops a user by deactivating or concluding a given user's enrollment. The process then deletes the user's WiW account, changes their status in CTL online to deferred or dropout, deletes their Canvas account (if they are a second-time melt), and sends them one of four emails depending on whether they are a first or second-time melt and whether they are a force or voluntary melt. The process also grades the student's final exam as "0" in order for the process not to double-melt users who have already been melted for a given course. The process always checks to see if a given user's grade for the final exam in a given course is "0" before counting the course as an additional melt.
#### Where should the reviewer start?
jobs/scheduling/dropMeltedUsers.js and jobs/scheduling/helpers/updateCanvas.js. You can also check out the new email templates in email_templates, and the new tests in test. 
#### How should this be manually tested?
In Sublime:

1. Add GLOBAL.KEYS and GLOBAL.CONFIG to the top of the dropMeltedUsers.js file:

  a) GLOBAL.KEYS = require('../../keys.js');
  b) GLOBAL.CONFIG = require('../../config.js');

2. Add GLOBAL.KEYS to the top of the WiWCCApi.js file: GLOBAL.KEYS = require('../keys.js');
3. Comment out the CronJob on line 18 of dropMeltedUsers.js.
4. Add a call to meltUsers() below the meltUsers function in dropMeltedUsers.js.

In WiW:

1. Create a fake user in the WiW test location:

  a) Go to https://app.wheniwork.com/employees/
  b) Add an employee.
  c) Assign your employee to the test location, and only the test location.

In Canvas:

1. Create a fake user in Canvas witih the same email address as your WiW user and assign them to a course:

  a) Go to https://crisistextline.instructure.com/accounts/1
  b) Add a user.
  c) Go to a course that currently has the "available" status (in other words, courses that show up on the course list at the above URL).
  d) Add your user to the course.
  e) Once your user shows up on the list of students in the course, you'll need to either deactivate or conclude their enrollment. Canvas provides you with ways to do either of these. You can go to the user's full details page: https://crisistextline.instructure.com/courses/:courseID/users/:userID to conclude their enrollment, or just go to the far right on the list of people in the course: https://crisistextline.instructure.com/courses/:courseID/users, click on the gear icon, and select "deactivate user." Any questions, just ask me.

In CTL Online:

[...]

PR #199, opened 2016-09-23
Revert "Modified dropMeltedUsers and updateCanvas to distinguish between firs…"

tongxiang:
Reverts CrisisTextLine/Elmer#191


tongxiang:
reverted because: 

https://admin.crisistextline.org/jira/browse/INT-401

PR #204, opened 2016-10-04
first draft of the cache

tongxiang:
#### What's this PR do?

https://admin.crisistextline.org/jira/browse/INT-413

1. Write a function in GLOBAL.CONFIG--accepting the argument of a user email--that creates an in-memory cache for the all users response object. (Making sure that it's a valid JSOn object)
2. If there's no object, it retrieves the users response object
3. If the user response object exists already, and if the user email doesn't exist in the user response object, it retrieves the user response again and attaches it to the variable. Then it returns the user object associated with the email. 
4. If the user email exists in the response object, it returns the appropriate user object. 

Then, replace all our "get all user" API calls with this function.

### DIFF
--- a/cache.js
+++ b/cache.js
@@ -1,4 +1,4 @@
-global.CONFIG = require('./config.js');
+'use strict';
 const api = CONFIG.WhenIWork;
 
 //cache for all WhenIWork users, active and deleted
@@ -6,7 +6,12 @@ function WiWUsersCache(){
   return new Promise (function(resolve, reject){
     api.get('users', {show_deleted: true}, function(response){
       if(response.message) reject('Call to get all WiW users failed');
-      resolve(response.users);
+      try {
+        if(typeof response !=='object') throw('WiW all users response is not in the proper JSON format');
+        resolve(response.users);
+      } catch (error) {
+        reject(error);
+      }
     });
   });
 };
--- a/cache.js
+++ b/cache.js
@@ -1,0 +1,37 @@
+global.CONFIG = require('./config.js');
+const api = CONFIG.WhenIWork;
+var users = {};
+var cache;
+
+//cache for all WhenIWork users, active and deleted
+(function WiWUsersCache(){
+
+
+    function singleton(){  
+      console.log('in singleton')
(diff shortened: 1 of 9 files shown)


jobs/scheduling/removePending.js
```diff
@@ -8,10 +8,10 @@ const api = CONFIG.WhenIWork;
 const CronJob = require('cron').CronJob;
 
 new CronJob(CONFIG.time_interval.pending_users, go, null, true);
-go();
+go(); 
 // get this party started
 function go(){
-	getPendingUsers()
+	getPendingUsers();
```
tongxiang: I think this is going to break, right?

mbogor: yes, fixed. thank you

cache.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
+const api = CONFIG.WhenIWork;
+
+//cache for all WhenIWork users, active and deleted
+function WiWUsersCache(){
+  return new Promise (function(resolve, reject){
+    api.get('users', {show_deleted: true}, function(response){
+      if(response.message) reject('Call to get all WiW users failed');
+      try {
+        if(typeof response !=='object') throw('WiW all users response is not in the proper JSON format');
```
tongxiang: Something can be an object but still not JSOn valid

I would make sure to JSOn.parse here

tongxiang: oh wait we added `JSON.parse` call to wheniwork-unofficial

we're good I think

runJobs.js
```diff
@@ -8,4 +9,8 @@ if (process.env.NODE_ENV !== 'production') {
   CONFIG.locationID.makeup_and_extra_shifts = CONFIG.locationID.test2;
 }
 
-var jobs = require('require-all')(__dirname + '/jobs');
\ No newline at end of file
+cache()
+.then(function(users){
+	global.cache= users;
+	var jobs = require('require-all')(__dirname + '/jobs');
```
tongxiang: const

tongxiang:
👍

PR #218, opened 2016-10-20
Melted data dump to mySQL

mikehow1984:
#### What's this PR do?

Expands on the meltedDataDumpToSlack.js script and appends a method to take data on each user's last completed checkpoint and dump it into CSV files

`/jobs/training/meltedDataDumpToCSV.js`

Run locally by adding this code after the require declarations in meltedDataDumpToSlack.js and meltedDataDumpToCSV.js:

``` javascript
//delete when pushing to Git
const KEYS = require('./keys.js');
var CONSOLE_WITH_TIME = function(msg){
  console.log(msg);
};

//end of delete
```

and comment out the cronjob in meltedDataDumpToCSV.js:

``` javascript
//Posts melted user data to #training channel on Slack every Wednesday at 10AM
//new CronJob(CONFIG.time_interval.melted_users_on_slack_and_csv_cron_job_string, function(){
 //   postMeltedUserDataToSlackAndCSV();
 // }, null, true);
```

also, comment out the `notifySlack` method, (unless you want to post info to slack, then I suggest changing the `SLACK_CHANNEL` variable to `#test` and checking that channel for the output)

``` javascript
function postMeltedUserDataToSlack(){
    return getCohortNumbers()
        .then(getEnrollmentsFromCohort)
        .then(getTotalAcceptedIntoTraining)
        .then(getGradAndFirstShiftCheckpointIDsForEachCohort) 
        .then(getTotalUsersWhoTookFirstShiftAndGraduated)   
        //.then(notifySlack)
        .catch(errorHandler);
}
```

CSV files are dumped in `/jobs/training/CSVdump`

https://admin.crisistextline.org/jira/browse/INT-381


jobs/training/meltedDataDumpToCSV.js
```diff
@@ -0,0 +1,267 @@
...
+			console.error('Error: ', err)
+		});
+}
+//obtains checkpoints from every class
+function getCheckpointIDsFromClasses(cohorts){
+	return new Promise(function(resolve, reject){
+		var start = Date.now();
+		CONSOLE_WITH_TIME("Getting checkpoint ID's from classes");
+		var cohortKeys = Object.keys(cohorts);
+		var assignmentPromises = [];
```
apackin: Don't need to declare this

jobs/training/meltedDataDumpToCSV.js
```diff
@@ -0,0 +1,267 @@
...
+			cohorts[cohortNum].class_ids.forEach(function(classNum){
+				urls.push({
+					url: `courses/${classNum}/assignments?per_page=1000`,
+					course: classNum,
+					cohort: cohortNum
+				});
+			});
+		});
+		
+		throttledURLCalls(urls, APICallsPerSecond, checkpointIDAPICall)
```
apackin: This could be a separate call in postMeltedUserDataToSlackAndCSV

jobs/training/meltedDataDumpToCSV.js
```diff
@@ -0,0 +1,267 @@
...
+		.then(putUserCohortInfoInCSVFile)
+		.catch(function(err){
+			console.error('Error: ', err)
+		});
+}
+//obtains checkpoints from every class
+function getCheckpointIDsFromClasses(cohorts){
+	return new Promise(function(resolve, reject){
+		var start = Date.now();
+		CONSOLE_WITH_TIME("Getting checkpoint ID's from classes");
```
apackin: Would prefer doing this after with meaningful data attached

jobs/training/meltedDataDumpToCSV.js
```diff
@@ -0,0 +1,267 @@
...
+}
+
+//request throttler for API calls
+function throttledURLCalls(urls, reqPerSecond, callback){
+	return new Promise(function(resolve,reject){
+		var intervalValue = 1000/reqPerSecond;
+		var requestCollector = [];
+		var count = 0;
+		
+		var go = setInterval(function(){
```
apackin: This is an interesting way to loop. Let's talk through it

jobs/training/meltedDataDumpToCSV.js
mikehow1984: assignmentPromises right? Yeah I used assignmentPromises in that scope, changed it, and forgot to delete this. Good catch.

jobs/training/meltedDataDumpToCSV.js
```diff
@@ -0,0 +1,267 @@
...
+			});
+		});
+		
+		throttledURLCalls(urls, APICallsPerSecond, checkpointIDAPICall)
+			.then(function(assignmentPromises){
+				var assignmentPromiseResponses = [];
+				Promise.all(assignmentPromises)
+					.then(function(res){
+						assignmentPromises.forEach(function(assignPromise){
+							var addAssignmentInfoToCohortInfo = assignPromise.then(function(assignmentInfo){
```
apackin: This looks like a .map?

jobs/training/meltedDataDumpToCSV.js
```diff
@@ -0,0 +1,267 @@
...
+								assignmentInfo.forEach(function(assignObj){
```
apackin: This could be its own function?

jobs/training/meltedDataDumpToCSV.js
```diff
@@ -0,0 +1,267 @@
...
+			});
+			return assignChanges;
+		}).catch(function(err){
+			console.error(`Problem getting grade change info for ${urlObj.url}: ${err}`);
+		});
+}
+
+//request throttler for API calls
+function throttledURLCalls(urls, reqPerSecond, callback){
+	return new Promise(function(resolve,reject){
```
apackin: const returnThing = urls.map(function(urlObj) {
			setTimeout(callback(urlObj), intervalValue);
		});

		resolve(returnThing);

mikehow1984: ultimately went with bluebird promises for this and set the concurrency. Works fantastically and speeds up the code.

jobs/training/meltedDataDumpToCSV.js
mikehow1984: refactored the code and I think now it makes more sense as a forEach.

jobs/training/meltedDataDumpToCSV.js
```diff
@@ -0,0 +1,237 @@
...
+			cohorts[cohortNum].assignmentNamesAndIDs = [];
+			cohorts[cohortNum].class_ids.forEach(function(classNum){
+				urls.push({
+					url: `courses/${classNum}/assignments?per_page=100`,
+					course: classNum,
+					cohort: cohortNum
+				});
+			});
+		});
+		resolve({cohorts: cohorts, urls: urls});
```
apackin: In ES6 the shorthand for this is {cohorts: cohorts, urls: urls} is {cohorts, urls}

jobs/training/meltedDataDumpToCSV.js
```diff
@@ -0,0 +1,237 @@
...
+	});
+}
+//obtains checkpoints from every class
+function obtainAssignmentNamesAndIDs(cohortObj){
```
apackin: You can combine ``` var cohorts = cohortObj.cohorts; var urls = cohortObj.urls; ``` into this step by destructuring the param as ```function obtainAssignmentNamesAndIDs({cohorts, urls})```

jobs/training/meltedDataDumpToCSV.js
```diff
@@ -0,0 +1,237 @@
...
+		.then(putUserCohortInfoInCSVFile)
+		.catch(function(err){
+			console.error('Error: ', err)
+		});
+}
+//creates an array of objects containing a url for each checkpoint ID API call, and the corresponding course and cohort number 
+function constructUrlsForCheckpointIdApiCalls(cohorts){
+	return new Promise(function(resolve, reject){
+		var cohortKeys = Object.keys(cohorts);
+		var urls =[];
```
apackin: Need to const cohorts = {}?

jobs/training/meltedDataDumpToCSV.js
```diff
@@ -0,0 +1,237 @@
...
+		});
+	}, {concurrency: numberOfConcurrentAPICalls});
+}
+
+//API call to grade changes that parses relevant info for CSV file (add more parameters if you want more columns for the CSV file)
+function getGradeChangesForAssignmentAPICall(urls){
+	return BBPromise.map(urls, function(urlObj){
+		return request(urlObj.url, 'Canvas')
+			.then(function(assignChange){
+				var assignChanges = [];
```
apackin: Creating an array from an array. This is a .map.

jobs/training/meltedDataDumpToCSV.js
```diff
@@ -0,0 +1,237 @@
...
+		CONSOLE_WITH_TIME('Getting last complete checkpoint for each user...');
+		var cohorts = cohortObj.cohorts;
+		var urls = cohortObj.urls;
+		getGradeChangesForAssignmentAPICall(urls)
+			.then(function(resolvedGradeChangeInfoRequests){
+				resolvedGradeChangeInfoRequests.forEach(function(request){
+					request.forEach(function(changeObj){
+						cohorts[changeObj.cohort].users.filter(function(user){
+							return changeObj.user_id == user.user_id;
+						}).map(function(user){
```
mikehow1984: This could just be a forEach. It basically checks each grade change, matches the student by id, then checks to see if it's the students' latest grade change. This is how you get the last completed checkpoint for the student.

tongxiang:
I would specify somewhere that `meltedDataDumpToSlack` is no longer a job that's running, but instead it's required by `meltedDataDumpToCSV.`

Perhaps in comments at the top of both files? 

When we consider the hierarchy of directory structures, each file should be its own job. Otherwise, it could be a good idea to put the `meltedDataDumpToSlack` in its own 'helper functions' file. Up to you.

tongxiang:
We also need to build the CSV dump as a route that can be hit, instead of a job that dumps to the directory. 

(Check out: `www/data/index.js`)

tongxiang:
👍

PR #223, opened 2016-10-31
Route was unable to find canvas.js

apackin:
#### What's this PR do?
Use CONFIG.root_dir  instead of relative path navigation
https://admin.crisistextline.org/jira/secure/RapidBoard.jspa?rapidView=5&view=detail&selectedIssue=INT-444


mbogor:
👍

PR #244, opened 2016-11-07
Add check for cache before running login route

apackin:
#### What's this PR do?
Checks for the user cache before running login script that depends on it. 
Fixes this error:

Stack trace
TypeError: Cannot convert undefined or null to object
TypeError: Cannot convert undefined or null to object
at allUsers (/var/www/elmer/www/scheduling/index.js:101:67)
at checkUser (/var/www/elmer/www/scheduling/index.js:157:3)


mbogor:
👍

PR #270, opened 2016-12-07
added workaround for Google Prediction API calls failing

mikehow1984:
#### What's this PR do?
Reverts Sherlock to Typeform grading if API calls to Google Prediction API fails
In the '/typeformAssignment/:assignment' route, comment out everything below ````const emailBody = submissionHTMLandGradeModifier.submissionHTML;```` in line 58 of ````index.js```` and add ````res.send(emailBody)````.

hit the route in Postman. 
This is a workaround for the current issue where models are not being returned by Google Prediction API, but a good permanent fix in case this problem reoccurs. 

https://admin.crisistextline.org/jira/browse/INT-514
#### Questions:
Perhaps I should add a method to email someone when this happens? Currently it outputs a warning message in console, but this is probably something we should be alerted to if it happens.


www/canvas/helpers.js
```diff
@@ -188,9 +188,19 @@ function getModelList(){
     predictionClient.getModels((err, models) => {
       if (err) CONSOLE_WITH_TIME(err);
```
apackin: Add the if (!models) here with the console logs and resolve([]);
That way the intended code flow is more clear for when it works correctly (same as before)

apackin: Just verifying - there are no issues if this resolves with an empty array, right?

mikehow1984: Yes, it resolves with an empty array, and since the models are called based on this array of model names, it'll just go with the Typeform grade for these questions because it assumes there was no Google model for them

PR #273, opened 2016-12-08
Email shifts route

mikehow1984:
#### What's this PR do?
creates a route that uses a user's email address to obtain their shifts
in ```/www``` run ```app.js```

hit the following route:
```localhost:3000/data/wiw/user-shifts?token=KEYS.dataRoute&email=USER_EMAIL```

replace **KEYS.dataRoute** with the actual key. **USER_EMAIL** is the person's email.
https://admin.crisistextline.org/jira/browse/INT-386

### CODE REVIEW

File: api_wiw/WiWRequests.js [hunk 1]
```diff
@@ -58,5 +66,27 @@ function mergeResponses (responses){
   return ({allUsers, allRequests});
 }
 
+function getUserIDFromEmail(email){
+  return new Promise(function(resolve, reject){
+    if(!email){
+      resolve();
+    } else{
+      api.get('users', function(response){
```
apackin: Should be able to use the global.USERS_CACHE of WiW users instead of making this big request -->
global.USERS_CACHE[email]

www/data/index.js
```diff
@@ -28,6 +28,21 @@ router.get('/wiw/timeoff', function (req, res) {
   });
 });
 
+router.get('/wiw/user-shifts', function (req, res) {
+  const email = req.query.email;
+    getShiftsFromEmail(email)
+      .then(shifts => {
+        console
```
apackin: ?

api_wiw/WiWRequests.js
```diff
@@ -58,5 +66,27 @@ function mergeResponses (responses){
...
+        resolve(response.users.filter(function(user){
+          return user.email === email;
```
apackin: Need to handle if the email is wrong/doesn't exist.

api_wiw/WiWRequests.js
```diff
@@ -58,5 +66,27 @@ function mergeResponses (responses){
   return ({allUsers, allRequests});
 }
 
+function getUserIDFromEmail(email){
+  return new Promise(function(resolve, reject){
+    if(!email){
+      resolve();
```
apackin: Doesn't hurt to have this here - but sending a req like this to /shifts is fairly limited - they would be better off using the tasks/shiftDump.js

apackin:
Works well - build in some error handling and you should be good to go. I'm wondering if we are better of redirecting them to shiftsDump if they send a generic request.

api_wiw/WiWRequests.js
mikehow1984: ah, so global.USERS_CACHE has all of these WiW objects living on it? Where is that constructed?

PR #284, opened 2017-01-02
Monitoring api

mbogor:
#### What's this PR do?
+ fixed a bug in graduation script, yes again
+ api checker route
forever www/app.js
then localhost:3000/monitoring/api

as for the graduation script, you can run it the usual way
https://admin.crisistextline.org/jira/browse/INT-466


jobs/training/pollCanvasForGraduatedUsersThenCreatePlatformAccount.js
```diff
@@ -101,7 +101,6 @@ function gradeLogs(courseID){
 }
 
 function getPlatformReadyId(data){
-  if (data.platformReadyId) resolve(data);
   return new Promise((resolve, reject)=>{
     canvas.getAssignments(data.courseID, CONFIG.canvas.assignments.platformReady)
```
apackin: If this is already a promise can we return it instead of wrapping it in a new promise?

www/monitoring/api.js
```diff
@@ -0,0 +1,97 @@
...
+		  res.render('monitoring/api', {data: str});
+	});
+}
+
+//check Canvas
+function pingCanvas(){
+	const data = {Canvas: null, WiW: null, VolunteerMatch: null, GoToWebinar: null};
+
+	return new Promise((resolve)=>{
+		canvas.getAllCourses('available')
```
apackin: This is already a promise -> can return this line

www/monitoring/api.js
```diff
@@ -0,0 +1,97 @@
...
+const router = express.Router();
+const canvas = require(CONFIG.root_dir + '/canvas.js');
+const helpers = require(`${CONFIG.root_dir}/www/canvas/helpers.js`);
+const WiW = CONFIG.WhenIWork;
+
+router.get('/', statusCheck);
+
+function statusCheck(req, res){
+
+	pingCanvas()
```
apackin: Do we need to do these in order or can we Promise.all() ?

www/monitoring/api.js
```diff
@@ -0,0 +1,97 @@
...
+			resolve(data);
+  	});
+	});
+}
+
+//check GoToWebinar
+function pingGTW(data){
+		var url = 'https://api.citrixonline.com/G2W/rest/organizers/'+KEYS.GTW.org_id+'/webinars';
+	var options = {
+      url:url,
```
apackin: es6 can be just `url,`

www/monitoring/api.js
```diff
@@ -0,0 +1,97 @@
...
+}
+
+//check Canvas
+function pingCanvas(){
+	const data = {Canvas: null, WiW: null, VolunteerMatch: null, GoToWebinar: null};
+
+	return new Promise((resolve)=>{
+		canvas.getAllCourses('available')
+		.then(response=>{
+			if (!response.length) data.Canvas = `All wrong ${response.body}`;
```
apackin: All wrong 😆

apackin:
Looks good 👍

jobs/training/pollCanvasForGraduatedUsersThenCreatePlatformAccount.js
mbogor: yep yep 100 correct will change thanks

www/monitoring/api.js
mbogor: we can i just like chains; is it any different functionality wise?

www/monitoring/api.js
mbogor: truth

www/monitoring/api.js
apackin: Functionally this way we are waiting until all of the previous pings have completed before moving on to the next one as opposed to sending them all out from the beginning and displaying the results when the slowest one gets back. But the promise.all fail-fast behavior may be an issue here...

PR #288, opened 2017-01-04
added punctuation checker to Sherlock

mikehow1984:
#### What's this PR do?
Handles punctuation errors in Sherlock
All major changes are in `www/canvas/helpers.js`
In the '/typeformAssignment/:assignment' route in `www/canvas/index.js`, comment out everything below ````const {gradeModifier, submissionHTML, comments} = responses[0];```` in line 58 until the `catch` statement and add ````res.send(submisstionHTML)})````.
https://admin.crisistextline.org/jira/browse/AT-20


www/canvas/helpers.js
```diff
@@ -242,4 +250,10 @@ function wordSplitter(answer){
   });
 }
 
+function punctuationHandler(str){
+  //checks for duplicate punctuation
+  const regCheck = /([!?():;'",<>\-%&$#@{}\\\/+=])\1{1,}|[.]{4,}|(?!\.\.)[!?.$#@%&]{2,}/g;
```
apackin: Should we also match '..' . I think there may be some rules about whether trainees are allowed to use ! - can you follow up with the training team about what grammar they want to enforce?

mikehow1984: Just contacted the training team and yes there is a "rule" (in that they are told to avoid them). I'll consult with Heather on the final say when she gets back tomorrow.

PR #294, opened 2017-01-17
Login rewrite

mbogor:
#### What's this PR do?
rewrites the WiW login script
www/scheduling/index.js
try to login as a new, existing and reactivated user
https://admin.crisistextline.org/jira/browse/AT-49


www/scheduling/index.js
```diff
@@ -11,147 +11,87 @@ const router = express.Router();
...
+});
 
-// reactivating a deleted user
-function reactivate(user){
+function reactivateUser(user, res){
   CONSOLE_WITH_TIME(`Reactivating user: ${user.email}`);
   return new Promise(function(resolve, reject){
-    api.update(`users/${user.id}`, {reactivate:true}, function(res){
-      if (res.message) reject(`User reactivation failed: ${res.message}`);
+    api.update(`users/${user.id}`, {reactivate:true}, function(response){
```
apackin: Can we call api => wiwApi?

www/scheduling/index.js
```diff
@@ -11,147 +11,87 @@ const router = express.Router();
...
+      if (response.message) reject(`User reactivation failed: ${res.message}`);
```
apackin: res.message => response.message

www/scheduling/index.js
```diff
@@ -11,147 +11,87 @@ const router = express.Router();
 router.use('/shifts', shiftSchedulingRouter);
 
 router.get('/login', function (req, res) {
-  CONSOLE_WITH_TIME('Login req query:', req.query);
+
+  var altEmail = helpers.generateAltEmail(req.query.email);
```
apackin: const

www/scheduling/index.js
```diff
@@ -11,147 +11,87 @@ const router = express.Router();
...
       else {
-        CONSOLE_WITH_TIME(`User reactivated ${res.user.email}`)
-        global.USERS_CACHE[res.user.email] = res.user;
-        delete global.USERS_CACHE.deleted[res.user.email];
-        resolve(res.user);
+        CONSOLE_WITH_TIME(`User reactivated ${response.user.email}`)
+        global.USERS_CACHE[response.user.email] = response.user;
+        delete global.USERS_CACHE.deleted[response.user.email];
+        WiWlogin(response.user, res);
+        resolve(response.user);
```
apackin: run WiWlogin(response.user, res) in the .then where this is called instead of starting a separate branch

www/scheduling/index.js
```diff
@@ -11,147 +11,87 @@ const router = express.Router();
...
-          CONSOLE_WITH_TIME(`Previously deleted user retrieved by cache: ${user.email}`);
-        })
-        .catch(function(error){
-          CONSOLE_WITH_TIME(`User ${users.deleted[email].id}reactivation failed: ${error}`)
-        });
-  // we didn't find the user in WiW records, we need to create a newUser
-    } else { callback(null, 'no user found');}
-}
+function createUser(req, res){
+  console.log(req.query)
```
apackin: -

www/scheduling/index.js
```diff
@@ -11,147 +11,87 @@ const router = express.Router();
...
+        activated: true,
+        locations: [CONFIG.locationID.regular_shifts, CONFIG.locationID.makeup_and_extra_shifts],
+        password: KEYS.wheniwork.default_password
+      };
 
-// creates a new user
-function createUser(newUser, callback){
-  CONSOLE_WITH_TIME(`Creating a new user ${newUser.email}`)
   return new Promise(function(resolve, reject){
     stathat.increment('Scheduling - Accounts Created', 1);
```
apackin: Should this be after the response?

www/scheduling/index.js
```diff
@@ -11,147 +11,87 @@ const router = express.Router();
...
     api.post('users', newUser, function (response) {
       if(!response || !response.user) {
         CONSOLE_WITH_TIME(`User ${newUser} creation failed: ${response}`)
-        reject();
+        reject(response.message);
       }
       else {      
         console.log(JSON.stringify(newUser));
```
apackin: console.logs here

www/scheduling/index.js
```diff
@@ -164,44 +104,85 @@ function createUser(newUser, callback){
         secondAPI.post('users/alerts', postBody, function () {});
         secondAPI.post('users/profile', {email: newUser.email}, function (profile) {
```
apackin: Do we care if these fail? let's handle this? Seems like setting up the alerts/profile should be its own function

www/scheduling/index.js
```diff
@@ -164,44 +104,85 @@ function createUser(newUser, callback){
...
   });
-}
+};
 
-// checks the status of the user: active, deleted, or newUser
+//update timezone
+function timezoneUpdate(user, timezone){
+  CONSOLE_WITH_TIME(`Timezone: ${timezone}, 'email: ${user.email}`);
+  // Parse notes
+  var notes = {};
```
apackin: let

www/scheduling/index.js
```diff
@@ -164,44 +104,85 @@ function createUser(newUser, callback){
...
+  try {
+    if(user.notes) notes = JSON.parse(user.notes);
+  } catch (e) {
+    if (user.notes !== undefined && user.notes.trim() !== '') {
+      notes[user.notes.trim()] = true;
+    }
+  }
+  notes.timezoneSet = true;
+  // Update the profile to reflect that they set their timezone
+  api.update('users/' + user.id, {notes: JSON.stringify(notes), timezone_id: timezone}, function(response){
```
apackin: can this fail? if so how should we handle?

www/scheduling/index.js
```diff
@@ -164,44 +104,85 @@ function createUser(newUser, callback){
         secondAPI.post('users/alerts', postBody, function () {});
         secondAPI.post('users/profile', {email: newUser.email}, function (profile) {
           CONSOLE_WITH_TIME(`User Successfully created ${profile.user.email}`);
-          callback(profile.user);
         });
+        WiWlogin(newUser, res);
```
apackin: The new promise should resolve and this should be taken from there

www/scheduling/index.js
```diff
@@ -164,44 +104,85 @@ function createUser(newUser, callback){
...
       }
-
     })
     .catch(function(error){
```
apackin: this .catch is on the callback?

apackin:
Nice improvement! Definitely a much simper, more understandable flow. 
There are a couple of code blocks that could me modularized (are performing separate functionality) and a few places where we are letting callbacks fail silently and not handling them - are we ok with that? 
The new Promises here are being called like functions that generate side effects instead of using a return / resolve. Keeping our code functional and using returns/resolve+thens will make debugging easier in the future and writing tests for this more feasible.

www/scheduling/index.js
mbogor: done

www/scheduling/index.js
mbogor: ty

www/scheduling/index.js
mbogor: 👍

www/scheduling/index.js
mbogor: 👍

www/scheduling/index.js
mbogor: good point

www/scheduling/index.js
mbogor: 👍

www/scheduling/index.js
mbogor: 👍

www/scheduling/index.js
mbogor: lol fixed

PR #299, opened 2017-01-23
Sherlock word count tweak

mikehow1984:
#### What's this PR do?
The minimum word count for assignment short answers was set to 10, even though Mini Roleplay and Full Roleplay is stated to be 15. This fixes that.

1. console.log out the `minWordCount` variable in the `formatHTML` function in `helpers.js`. 
2. Run `node app` in `www`.
3.
[full-roleplay.txt](https://github.com/CrisisTextLine/Elmer/files/725057/full-roleplay.txt)
[mini-roleplay.txt](https://github.com/CrisisTextLine/Elmer/files/725058/mini-roleplay.txt)
These are requests for the `canvas/typeformAssignment/Mini Roleplay` and `canvas/typeformAssignment/Full Roleplay` routes. Put these in Postman and submit.
4. `minWordCount` should equal 15 for either assignment.
https://admin.crisistextline.org/jira/browse/AT-63


www/canvas/index.js
```diff
@@ -73,6 +72,7 @@ router.post('/typeformAssignment/:assignment', function (req, res) {
     })
     .then(() => res.send({message: `Successfully submitted ${assignment} with ${JSON.stringify(emailAndAnswers)}`}))
     .catch((err) => {
+      console.err(err);
```
mbogor: we're trying to switch to throwing new Error(message), so that in the catch blocks we can console log error.stack which gives us a better idea of what the error is and where it occurred. other than that great PR

PR #306, opened 2017-01-30
Penalize trainees for asking too many questions on an assignment

mikehow1984:
#### What's this PR do?
In the GCT Skill Bite, RA Skill Bite, Mini Roleplay, and Full Roleplay assignments, we want to be able to penalize trainees for asking too many questions. for each assignment there is a different scoring metric defined in `config.js` under `CONFIG.sherlock.questionPenalty`

I've also refactored and cleaned up Sherlock in preparation for Sherlock V3
`www/canvas/index.js`
In the `/typeformAssignment/:assignment` route in the `helpers.predictEachQuestion` promise chain, in `.then(responses =>`, comment out everything in the `.then` method under the `const {gradeModifier, submissionHTML, comments} = responses[0];` line and add `res.send(submissionHTML);})`

go to Postman and use the following responses for each route:

`localhost:3000/canvas/typeformAssignment/Good-Contact Techniques Skill Bite`:
[GCT-response.txt](https://github.com/CrisisTextLine/Elmer/files/739676/GCT-response.txt)

`localhost:3000/canvas/typeformAssignment/Full Roleplay`:
[full-roleplay.txt](https://github.com/CrisisTextLine/Elmer/files/739677/full-roleplay.txt)

`localhost:3000/canvas/typeformAssignment/Mini Roleplay`:
[mini-roleplay.txt](https://github.com/CrisisTextLine/Elmer/files/739678/mini-roleplay.txt)

To go over the question threshold, add questions to short-text answers (or just add question marks after each word in a question)

https://admin.crisistextline.org/jira/browse/AT-68


www/canvas/sherlock/googlePredictionHelpers.js
```diff
@@ -0,0 +1,100 @@
+'use strict';
```
apackin: Updated Airbnb style guide don't need 'use strict' in sub modules ^^

www/canvas/helpers.js
```diff
@@ -57,9 +35,11 @@ helpers.sortQuestions = function(formResponse){
 //Uses Google Prediction API and processes the score info
 helpers.predictEachQuestion = function(sortedQuestions, answers, assignment){
   if(CONFIG.predictions[assignment]){
-    return useGooglePredictionScoring(sortedQuestions, answers, assignment)
-          .then(processGooglePredictionScoring)
-          .catch(errorHandler);
+    return Predict.useGooglePredictionScoring(sortedQuestions, answers, assignment)
+          .then(Predict.processGooglePredictionScoring)
+          .catch(function(err){
+            CONSOLE_WITH_TIME('trace', `Something with wrong using Google Prediction for ${assignment}\n${err.trace}`);
```
apackin: We've been using err.stack, how does err.trace compare to that?

www/canvas/sherlock/googlePredictionHelpers.js
```diff
@@ -0,0 +1,100 @@
...
+function getModelList(){
+  return new Promise(function(resolve,reject){
+    predictionClient.getModels((err, models) => {
+      if (err) CONSOLE_WITH_TIME(err);
+      // `models` is an array of Model objects.
+      if(!models){
+        //added to bypass Google Prediction if API call for models fails
+        CONSOLE_WITH_TIME('WARNING: Call to get models from Google Prediction failed');
+        CONSOLE_WITH_TIME('Reverting to using normal Typeform grading.');
+        resolve([]);
```
apackin: Do we want to reject here instead of resolving empty?

www/canvas/sherlock/graderMethods.js
```diff
@@ -0,0 +1,104 @@
...
+  }
+};
+
+//checks if trainee got an answer wrong by checking the next question. 
+Grade.typeformGrade = function(questionNum, nextQuesTitle){
+  const regex = /\d{2}[a-zA-Z]./;
+  const fBracketRegex = /\[\[/;
+  const bBracketRegex = /\]\]/;
+  const numberLetter = nextQuesTitle.match(regex);
+  // fBracketRegex.test(nextQuesTitle)
```
apackin: Are we keeping this comment?

www/canvas/sherlock/graderMethods.js
```diff
@@ -0,0 +1,104 @@
...
+Grade.questionCounter = function(text){
+  const quesReg = /\b\?{1}/g;
+  const match = text.match(quesReg);
+  return match ? match.length : 0;
+};
+
+//determines the penalty for too many questions in an assignment
+Grade.questionPenalty = function(numOfQuestions, assignment){
+  let penalty = 0;
+  if(CONFIG.sherlock.questionPenalty[assignment]){
```
apackin: I think it would make more sense for questionPenalty[assignment] to be an object that matches the # of questions asked to the penalty. Using two arrays that match on idx is not recommended

www/canvas/sherlock/graderMethods.js
```diff
@@ -0,0 +1,104 @@
...
+  }
+  punctErrors = punctErrors.map(function(pErr){
+    return `\"${pErr}\"`;
+  });
+  return punctErrors;
+};
+
+//counts number of questions in assignment
+Grade.questionCounter = function(text){
+  const quesReg = /\b\?{1}/g;
```
apackin: Multiple '?'s in the same response count as multiple questions in the overall quiz? Just double checking that is the intention of training team.

mikehow1984: Good question. My understanding was that it was just the number of questions in the entire assignment, but I'll definitely double check.

www/canvas/sherlock/graderMethods.js
mikehow1984: I meant to ask you what you thought the best implementation of this was. I went this route because the scores depend on a range of numbers and objects would be a bit clunky. We should discuss a bit.

www/canvas/sherlock/graderMethods.js
mikehow1984: No, removing this.

apackin:
Smart refactoring! This will definitely make the code more maintainable in the future. 
I didn't want to drive you crazy with comments about spacing/syntax/es6 so I committed those changes https://github.com/CrisisTextLine/Elmer/pull/306/commits/01d8e7614d999605d1b2bbef2149e44981ec558c . Definitely check them out and let me know if you have an questions/comments/concerns.
As far as the added functionality, I left comments in your original commit https://github.com/CrisisTextLine/Elmer/pull/306/commits/c5981e1a6c988607d4b3c5a0a426c8dcff9b1e86 . I don't feel right about using the arrays coordinated by idx in questionPenalty, but I understand why you went that route - can we try to come up with a better solution?

www/canvas/helpers.js
mikehow1984: This is a typo. It should be err.stack.

www/canvas/sherlock/googlePredictionHelpers.js
mikehow1984: This is an older change which made more sense when Sherlock V2 was first implemented, but now that more functionality has been added and we're moving away from Typeform vanilla grading, this should definitely fail instead of reverting to Typeform.

www/canvas/helpers.js
```diff
@@ -45,245 +22,104 @@ helpers.extractEmailAndFormatSubmission = function(formResponse) {
...
-          return model.id;
-        });
-        resolve(modelList);
-      }
-    });
-  });
-}
+  let submissionHTML = `<p> ${pairedQtoA.join('</p><br><p>')} </p>`;
+  if (quesPenalty) {
+    gradeModifier += quesPenalty;
```
apackin: Should this be -=?

mikehow1984: Yes, excellent catch

PR #311, opened 2017-02-13
Sherlock V3 Feedback Prediction

mikehow1984:
#### What's this PR do?
This adds predicted constructive feedback to relevant questions in the the Good Contact Techniques Canvas assignment.
go to `www` and run `node app`
In Postman, hit the route `localhost:3000/canvas/typeformAssignment/Good-Contact Techniques Skill Bite` with a valid Typeform request body. You can use this one:
[reqObjectTypeform.txt](https://github.com/CrisisTextLine/Elmer/files/771909/reqObjectTypeform.txt)
https://admin.crisistextline.org/jira/browse/AT-201
https://admin.crisistextline.org/jira/browse/AT-231


www/canvas/sherlock/googlePredictionHelpers.js
```diff
@@ -60,22 +68,73 @@ function getModelList() {
...
   });
 }
 
+function predictCommentModel(modelID, questionObj, comment) {
+  return new Promise((resolve, reject) => {
+  	//have the model ID include the "-Comment-${comment}"
+    predictionClient.model(modelID).query(questionObj.answer, (err, result) => {
+      if (err) {
+      	CONSOLE_WITH_TIME(`Error predicting ${modelID}\n${err.stack}`);
+        reject();
```
apackin: Be careful using rejects in these -
If you promise.all them and any of them rejects then the whole Promise.all rejects.

www/canvas/sherlock/googlePredictionHelpers.js
```diff
@@ -8,45 +9,52 @@ const predictionClient = prediction({
...
-    });
-  const gradeModifier = scoreModifiers.reduce((a, b) => a + b, 0);
-  return ({ gradeModifier, sherlockComments });
-};
+Predict.processGooglePredictionScoring = ({ predictions, reqObj }) => 
+	predictions.then((predict) => {
+		predict.forEach((p) => {
+			if (p.prediction_type === 'main') {
+				let questionIdx;
+				const question = reqObj.sortedQuestions.filter((q, idx) => {
```
apackin: Looks like you want .find in these cases

apackin: If you're using the question const in the else case you may want to do it in the larger scope

www/canvas/sherlock/googlePredictionHelpers.js
```diff
@@ -8,45 +9,52 @@ const predictionClient = prediction({
...
+        	1;
+				reqObj.sherlockGradeModifier += (p.predictedScore - typeformGrade);
+			} else if (p.prediction_type === 'comment') {
+				reqObj.sortedQuestions.filter((q) => {
+					return q.id === p.question_id;
+				})[0].feedback.push(p);
+			}
+		});
+		return reqObj;
+	}).catch((e) => CONSOLE_WITH_TIME(`Something happened while using Google Prediction API\n${e.stack}`));
```
apackin: may want to specify where this error is

www/canvas/sherlock/feedback.js
```diff
@@ -0,0 +1,33 @@
+const Feedback = {};
+
+function removeUnsureFeedback(feedback) {
+	return feedback.filter((f) => f.weight > 0.75);
+}
+
+function chooseFeedback(feedback) {
+	const negativeFeedback = feedback.filter((f) => f.predictedScore === 1);
+	const positiveFeedback = feedback.filter((f) => f.predictedScore === 0);
+	let ultimateFeedback = [];
```
apackin: :1st_place_medal:

www/canvas/index.js
```diff
@@ -35,52 +35,42 @@ router.put('/firstShift', function (req, res) {
...
-        if(CONFIG.predictions[assignment]) formResponse.calculated.score += gradeModifier;   
+  helpers.predictEachQuestion(request)
+    .then(helpers.formatFeedback)
+    .then((reqObject) => {
+      return Promise.all([ 
+        helpers.formatHTMLresponse(reqObject), 
+        canvas.comp.retrieveUserCourseAssignmentIds(reqObject.email, reqObject.assignment, {delay}, logErrEmailHeather)
+      ]);
+    }).then(responses => {
+      request = responses[0];
```
apackin: request

config.js
```diff
@@ -187,8 +187,70 @@ CONFIG.sherlock = {
       18: 1.5,
       16: 1,
       14: .5,
-    }
-  }
+    },
+  },
+  feedback: {
```
mikehow1984: Yes, along with updating the data, Heather will give me the actual values for this

www/canvas/sherlock/googlePredictionHelpers.js
```diff
@@ -1,5 +1,6 @@
 const Grade = require('./graderMethods.js');
 const prediction = require('@google-cloud/prediction');
+const Bluebird = require('bluebird');
```
mikehow1984: I used it to throttle calls to Google Prediction originally, then went another route and didn't delete the Bluebird declaration. Deleting.

www/canvas/sherlock/googlePredictionHelpers.js
mikehow1984: Good call. It was originally written for the main question score prediction, and you'd probably want everything to reject if one rejects. With the comments, you could let one or two reject without compromising much and we wouldn't want a bad response for a comment that may not even be used to stop everything.

www/canvas/sherlock/googlePredictionHelpers.js
```diff
@@ -60,22 +68,73 @@ function getModelList() {
...
+          question_id: questionObj.id,
+          prediction_type: 'comment',
+        });
+      }
+    });
+  });
+}
+
+function modelpredictThrottler(arr) {
+	const delay = arr.length < 100 ? 100 : 1000;
```
mikehow1984: As it stands now, it takes about 8 seconds. I added this throttler because I thought that Google Prediction calls were failing because of them hitting a limit, which wasn't the case. I kept the throttler in just in case it was needed in the future, but looking at this line of code, adding a second between calls may be a bit overkill. I'm removing the ternary operator and just making the delay equal 100 for now.

www/canvas/sherlock/googlePredictionHelpers.js
```diff
@@ -60,22 +68,73 @@ function getModelList() {
...
+	function* iterGen(array) {
+    let i = 0;
+    while(i !== array.length) {
+    	yield array[i];
+    	i++;
+    }
+    return { next: () => { done: true } };
+	}
+
+	return new Promise((resolve, reject) => {
```
mikehow1984: I was actually looking for something like mapSeries in the Bluebird library. This is much easier to read.

mikehow1984: The generator method seems to be about 2.5 times quicker just because it allows for a little bit of concurrency while also delaying the time between each API call, while the mapSeries/delay method waits for the last call to complete, delays, and then executes a new call. I think for now readability trumps speed because it's a difference between 9 seconds and 24 seconds. if we decide to add more Google Prediction in a single assignment, we may want to revisit the generator method.

mikehow1984: I think I'm going to keep the generator throttler in the code commented out in case we want to switch it out at some point.

www/canvas/helpers.js
```diff
@@ -98,37 +101,54 @@ helpers.formatHTMLresponse = (reqObject) => {
...
+  const rawVars = question.match(/{{([^}}]+)}}/g);
+  let newQuestion = question;
+  if(rawVars){
+    newQuestion = rawVars.reduce((acc, val) => {
+      const variable = val.slice(2,-2).split("_");
+      if(variable[0] === 'answer') {
+        return acc.replace(val, questionObjs.filter((obj) => 
+          variable[1] === obj.id)[0].answer);
+      }
+      if(variable[0] === 'var' && variable[1] === 'score') return acc.replace(val, score);
```
apackin: Can we leave this as 'var_score' and match on that?

www/canvas/helpers.js
```diff
@@ -98,37 +101,54 @@ helpers.formatHTMLresponse = (reqObject) => {
...
+  submission.trueGrade = submission.typeformScore ? submission.typeformScore + submission.sherlockGradeModifier : null;
+  return submission;
 };
 
+function rawVariableScrubber(question, questionObjs, score) {
+  const rawVars = question.match(/{{([^}}]+)}}/g);
+  let newQuestion = question;
+  if(rawVars){
+    newQuestion = rawVars.reduce((acc, val) => {
+      const variable = val.slice(2,-2).split("_");
```
apackin: I think you can drop the slice and declare this as index 0, then match on the whole term

config.js
```diff
@@ -191,66 +191,93 @@ CONFIG.sherlock = {
...
+      ranking: 1,
+      penalty: Infinity,
     },
     'Giving Advice': {
-      positive: "+ Didn't give advice. (We don't want to give advice to texters)",
-      negative: "- Gave advice. (We don't want to give advice to texters)",
+      positive: `• Nondirective. Great job not giving advice to your texter! It’s hard, especially in cases where they ask for our thoughts. But encouraging them to brainstorm their own solutions will help them better deal with their next moment of crisis.`,
+      negative: `• Directive. We want to avoid being directive — telling a texter what to do or feel. For this roleplay, there was room for you to let the texter lead a little more. We avoid telling a texter they should/shouldn’t feel something and we ask open questions that encourage them to brainstorm solutions. It gets them thinking and can usually give you something to use as a springboard to explore more of their situation.`,
+      ranking: 6,
+      penalty: Infinity,
```
apackin: Infinity? 😆

www/canvas/helpers.js
```diff
@@ -124,14 +120,17 @@ helpers.formatHTMLresponse = (submission) => {
...
-  submission.comments = comments.creator(commentTally, minWordCount);
   submission.submissionHTML = `<p> ${HTMLresponse.join('</p><br><p>')} </p>`;
   if (commentTally.questionPenalty) {
     submission.sherlockGradeModifier -= commentTally.questionPenalty;
     submission.submissionHTML += `<p><br><h3>Points deducted for asking texter too many questions: ` +
         `<b>Asked texter ${commentTally.numberOfQuestions} questions.</b></h3></p>`;
   }
-  submission.trueGrade = submission.typeformScore ? submission.typeformScore + submission.sherlockGradeModifier : null;
+  submission.trueGrade = submission.typeformScore + submission.sherlockGradeModifier;
+  if (submission.trueGrade < 0) submission.trueGrade = 0;
```
apackin: :1st_place_medal:

PR #332, opened 2017-03-30
AT-416: Sherlock V3 removed 'sent intro message' feedback model

mikehow1984:
#### What's this PR do?
"Sent Intro Message" model was acting misfiring. Instead of fixing it, we're moving it completely. This PR removes the model name/info from the config file so that Sherlock doesn't check for it.

-in `www/canvas/helpers.js` add `CONSOLE_WITH_TIME(submission.comments);` after line 133.

-Use this Typeform response in Postman (it results in no negative comments): 
[sherlock.txt](https://github.com/CrisisTextLine/Elmer/files/880390/sherlock.txt)

-run `node app` in the `www` folder

-send response

-???

-PROFIT
I like turtles.
https://admin.crisistextline.org/jira/browse/AT-416


mbogor:
👍

PR #329, opened 2017-03-27
Okta platform update

mbogor:
#### What's this PR do?
checks okta for user profile update and updates their platform accounts 
#### Where should the reviewer start?
only jobs/okta/platformUserUpdate.js
i branched from another okta branch because i needed the okta.js file, so the rest of the code in here has its own PR
#### What are the relevant tickets?
https://admin.crisistextline.org/jira/browse/AT-323

PR #334, opened 2017-04-03
Merge pull request #329 from CrisisTextLine/oktaPlatformUpdate

mbogor:
Okta platform update

#### What's this PR do?
adds salesforce fixes to the okta branch


jobs/okta/WiWShiftsAndPlatformAttendance.js
```diff
@@ -59,7 +61,7 @@ function getOktaPlatformUsers(timeOff){
 		const allUsers = {};
 		users.forEach(user=>{
 			const email = user.credentials.userName;
-			if (!timeOff[email]) allUsers[email] = {earlyChurn: true};
+			if (!timeOff[email]) allUsers[email] = {Two_Weeks_Off__c: true};
```
apackin: 'Two_Weeks_Off__c' <-- can we name this something else?

mbogor: it's the name of the property on the salesforce object that i think ramneet gave it. so i guess i can ask her but why?

apackin: I just felt like it's ambiguous. '__c' refers to CCs (not coaches). Two_Weeks_Off means they are off for two weeks without requesting time off which is more like Two_Weeks_Absentee. Not a big deal if it's already coded into salesforce, but thought it could be better

PR #338, opened 2017-04-06
platform auth updated

mbogor:
#### What's this PR do?
DO NOT MERGE BEFORE TC MERGES HIS UPDATED API BRANCH ON PLATFORM
MERGE AFTER VERBAL CONFORMATION ONLY
updates platform auth for the graduate trainees job
update your keys file before running any scripts requiring platform auth on your local
#### Questions:
should we create a platform file a la canvas.js? or should we just leave this since it's only two scripts that are pushing to platform?


jobs/training/pollCanvasForGraduatedUsersThenCreatePlatformAccount.js
```diff
@@ -171,6 +171,7 @@ function createEmail(student) {
     lastName: name[1],
     email: student.login_id,
     coach: student.coach,
+    key: KEYS.platform.apiKey
```
kidatsy: you define the keys variable here, but use the same `KEYS...` var below. Let's do one or the other?

jobs/training/pollCanvasForGraduatedUsersThenCreatePlatformAccount.js
```diff
@@ -180,15 +181,22 @@ function updatePlatform(body) {
...
     form: body
-  }, function (error, res) {
+  };
+  const timestamp = new Date();
+  const signature= crypto.createHash('sha256').update(KEYS.platform.apiKey+'|'+JSON.stringify({data:body})+'|'+KEYS.platform.apiSecret +'|'+ timestamp).digest('hex');
+  const token = JSON.stringify({
+      signature,
+      timestamp
+    });
+  params.body.token = token;
```
kidatsy: `params.form` here

PR #342, opened 2017-04-17
fixed bug where Sherlock wasn't correctly grading wordcount

mikehow1984:
#### What's this PR do?
Trainees were having points deducted for having less than 10 words in a response even if they had more than 10 words. This is because trainees were submitting answers without punctuation in the last sentence, and the regex check for sentences wasn't picking up the last sentence. This fixes the regex.

Take the `wordSplitter` function and test it with strings in repl.it. The case that causes a bug is a string with a sentence with correct punctuation that's less that 10 words, followed by a sentence with no punctuation. 
https://admin.crisistextline.org/jira/browse/AT-511


www/canvas/sherlock/graderMethods.js
```diff
@@ -24,9 +24,7 @@ Grade.wordSplitter = answer => {
     wordStats.wordCount += wordArr.length;
     const wordsToCheck = [];
     for (let i = 0; i < wordArr.length; i++) {
-      if (i === 0 || wordArr[i][0] !== wordArr[i][0].toUpperCase()) {
-        wordsToCheck.push(wordArr[i]);
-      }
+      wordsToCheck.push(wordArr[i]);
```
mikehow1984: I got rid of the `if` statement because I couldn't figure out why I had included it in the first place (and didn't realize that that made the `for` loop redundant. I realize it was so proper nouns wouldn't be spell checked, so I'm re-adding the `if`.

PR #357, opened 2017-05-11
Mike/melted user bug/at 533

mikehow1984:
#### What's this PR do?
-fixes dropVoluntaryMelter script bring broken due to referencing a non-existent object key.
-fixes `(var)[Symbol.iterator] is not a function` error that is caused due to an incompatibility between Bluebird promises and native promises (The `request-promise` package in node uses bluebird promises that were used in a native `Promise.all` function). 
https://admin.crisistextline.org/jira/browse/AT-533


mbogor:
great! good catch on the request library's bluebird promises

PR #368, opened 2017-05-25
Mike/turnoff forcemelt by cohort better implementation/at 609

mikehow1984:
#### What's this PR do?
adds better way to change which cohorts are melted and to turn off melting completely (adds flag in config file to add cohorts
https://admin.crisistextline.org/jira/browse/AT-609


jobs/training/autoForceMelt.js
```diff
@@ -28,9 +28,14 @@ new CronJob(CONFIG.time_interval.cronEverydayAt6AM, autoMeltCanvasUsers, null, t
...
 function autoMeltCanvasUsers() {
+  if(CONFIG.forceMelt.turnOffMelts) {
+    CONSOLE_WITH_TIME(`User melt scripts are turned off. Edit CONFIG file to turn it back on.`);
+    return;
+  }
   return canvas.getAllCourses('available')
   .then(courses => bluebird.map(courses, course => {
-    if (!/noMelt/i.test(course.sis_course_id)) {
+    const cohortNum = course.sis_course_id.match(/c\d+/)[0];
+    if (!/noMelt/i.test(course.sis_course_id) && !CONFIG.forceMelt.cohortsNotToMelt.includes(cohortNum)) {
```
apackin: || instead of && ?

jobs/training/dropVoluntaryMelters.js
```diff
@@ -19,8 +19,20 @@ new CronJob(CONFIG.time_interval.cron_twice_per_day, dropVoluntaryMelts, null, t
 // Find voluntarily melted users (completed enrollment in Canvas)
 // Modify their CTL online role, email them, delete their WiW and Canvas accounts as needed
 function dropVoluntaryMelts() {
+  if(CONFIG.forceMelt.turnOffMelts) {
```
apackin: Should we add the logic for !/noMelt/i.test(course.sis_course_id)  in here too?

mikehow1984:
adding "noMelt" to sis_course_id could break the way we get a coach's name from the sis_course_id (which can be fixed with another regex tweak). I'm also not sure that they can change the sis_course_id when the course is open (at least, Ramneet said she couldn't).

jobs/training/autoForceMelt.js
mikehow1984: _if_ sis_course_id does not have "nomelt" _and_ CONFIG.forceMelt.cohortsNotToMelt does not include the cohortNum

We want the melt script to run if both statements are true. If it were _or_ it's run if the cohort was tagged "noMelt" but it's not in the "cohortsNotToMelt" config

jobs/training/dropVoluntaryMelters.js
mikehow1984: 👍

apackin:
1 - TBH I think that we should be finding coaches names in a smarter way anyway 
2- updating sis_course_id should be possible (I know Heather has done it) let me look into it and get back to you

mikehow1984:
Agreed. Do you think we should keep the `CONFIG.forceMelt.turnOffMelts` flag as a quick way to stop the script from running?

jobs/training/autoForceMelt.js
apackin: haha, true - the context here is confusing

 but we're going to remove '&& !CONFIG.forceMelt.cohortsNotToMelt.includes(cohortNum)' right?

apackin:
I'm ambivalent about the `CONFIG.forceMelt.turnOffMelts` flag. If you think that will make things easier in the future feel free to keep it.

canvas.js
```diff
@@ -444,7 +444,33 @@ canvas.comp = {
       const context = `Error finding ${assignment} in course found for ${userEmail}`;
       errFunc(context, err);
     });
-  }
+  },
+  retrieveCoachInfoFromCourseId: (courseId) => {
+    return canvas.retrieveCourseEnrollment(courseId, ['completed', 'active', 'inactive'], 'TeacherEnrollment')
+      .then((coachDataResponse) => {
+        const coachObj = { name: null, email: null, courseId: courseId };
+        if (!coachDataResponse.length || !coachDataResponse[0].sis_course_id) return coachObj;
```
apackin: I think we would prefer this throw an error than return a null user?

canvas.js
```diff
@@ -444,7 +445,36 @@ canvas.comp = {
...
+          throw new Error(`Could not find the coach for course ${courseId}`);
+        }
+        const coachNameFromSisCourseId = coachDataResponse[0].sis_course_id ? 
+          coachDataResponse[0].sis_course_id.match(/c\d+(.+)/) :
+          null;
+        let coachData;
+        if(coachNameFromSisCourseId) {
+          const nameRegEx = new RegExp(coachNameFromSisCourseId[1],"i");
+          //checking coach name against name in sis_course_id because some courses have 2 coaches
+          coachData = coachDataResponse.filter(coach => {
```
apackin: .find is probably a better fit here -> https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find?v=control

canvas.js
```diff
@@ -444,7 +445,36 @@ canvas.comp = {
...
+  retrieveCoachInfoFromCourseId: (courseId) => {
+    return canvas.retrieveCourseEnrollment(courseId, ['completed', 'active', 'inactive'], 'TeacherEnrollment')
+      .then((coachDataResponse) => {
+        if (!coachDataResponse.length) {
+          throw new Error(`Could not find the coach for course ${courseId}`);
+        }
+        const coachNameFromSisCourseId = coachDataResponse[0].sis_course_id ? 
+          coachDataResponse[0].sis_course_id.match(/c\d+(.+)/) :
+          null;
+        let coachData;
```
apackin: let's set this to false

canvas.js
```diff
@@ -444,7 +445,36 @@ canvas.comp = {
...
+  },
+  retrieveCoachInfoFromCourseId: (courseId) => {
+    return canvas.retrieveCourseEnrollment(courseId, ['completed', 'active', 'inactive'], 'TeacherEnrollment')
+      .then((coachDataResponse) => {
+        if (!coachDataResponse.length) {
+          throw new Error(`Could not find the coach for course ${courseId}`);
+        }
+        const coachNameFromSisCourseId = coachDataResponse[0].sis_course_id ? 
+          coachDataResponse[0].sis_course_id.match(/c\d+(.+)/) :
+          null;
```
apackin: Would also use false here instead of null

PR #375, opened 2017-06-20
Mike/map coach emails to trainees/at 720

mikehow1984:
#### What's this PR do?
Instead of getting feedback emails from "support@crisistextline.org" trainees how get feedback from their coach
https://admin.crisistextline.org/jira/browse/AT-720


canvas.js
```diff
@@ -259,7 +259,10 @@ canvas.getActiveStudentsInCourse = (courseID, delay) => {
 //returns array of student's enrollment objects containing course and user data
 canvas.getEnrollment = function(userID, enrollmentState, delay){
   let url = `${KEYS.canvas.url}api/v1/users/${userID}/enrollments?per_page=100`;
-  if(enrollmentState) url += `&state[]=${enrollmentState}`;
+  if(enrollmentState) {
```
morrismatts: Nitpick, but I think our code style includes a space between `if` and the open paren.

canvas.js
```diff
@@ -259,7 +259,10 @@ canvas.getActiveStudentsInCourse = (courseID, delay) => {
...
+    if(enrollmentState.constructor !== Array) enrollmentState = [ enrollmentState ];
```
morrismatts: Same nitpick.

canvas.js
```diff
@@ -259,7 +259,10 @@ canvas.getActiveStudentsInCourse = (courseID, delay) => {
...
+    enrollmentState.forEach((state) => url += `&state[]=${state}`);
```
morrismatts: Where is this code running? I assume Elmer is running the same version of Node as the platform, which I thought didn't support ES6 syntax. It looks like we're already using it other places though, so maybe (hopefully!) I'm wrong.

canvas.js
```diff
@@ -273,12 +276,10 @@ canvas.getEnrollment = function(userID, enrollmentState, delay){
 //returns array of course's enrollment objects containing course and user data
 canvas.retrieveCourseEnrollment = function(courseID, enrollmentState, enrollmentType) {
   let url = `${KEYS.canvas.url}api/v1/courses/${courseID}/enrollments?per_page=100`;
-  if(enrollmentState){
+  if(enrollmentState) {
```
morrismatts: Add space after `if`.

canvas.js
```diff
@@ -469,8 +470,26 @@ canvas.comp = {
...
         };
       })
-      .catch(function(err){
-        CONSOLE_WITH_TIME(`Finding coach info for the course with ID ${courseId} in Canvas failed.`);
+      .catch((err) => {
+        CONSOLE_WITH_TIME('trace', `Finding coach info for the course with ID ${courseId} in Canvas failed: ${err.stack}`);
+      });
+  },
+  retrieveCourseEnrollmentIdFromEmail: (email) => {
+    return canvas.getUsers(email)
```
morrismatts: (Assuming we can use ES6) you can remove the `return` and either remove the curly brackets or switch them to parens.

canvas.js
```diff
@@ -469,8 +470,26 @@ canvas.comp = {
...
+      .then((users) => {
+        const userId = users.length ? users[0].id : undefined;
```
morrismatts: Does Elmer include lodash? If so, you can rewrite this as `_.get(users, '0.id')`.

canvas.js
```diff
@@ -469,8 +470,26 @@ canvas.comp = {
...
+        if (!userId) throw new Error(`There are no users named ${email}.`);
+          
+        return canvas.getEnrollment(userId, ['active', 'inactive', 'completed']);
+      })
+      .then((enrollments) => {
+        const courseId = enrollments.length ? enrollments[0].course_id : undefined;
+        if (!courseId) throw new Error(`User ${email} has no enrolled courses.`);
+
```
morrismatts: In a couple of spots you're adding empty lines before `return` statements. I think it's generally better to remove blank lines if they're not necessary. But if this is the style used in Elmer, feel free to ignore me.

www/canvas/helpers.js
```diff
@@ -68,6 +68,26 @@ helpers.getFullName = (submission) => {
   });
 };
 
+helpers.getCoachInfo = (submission) => {
+  return new Promise((resolve, reject) => {
```
morrismatts: Mentioned this above, but there are also a few spots in this file where you could drop the `return` and either remove the curly brackets or switch them to parens.

www/canvas/helpers.js
```diff
@@ -236,13 +256,12 @@ helpers.triggersBasedOnAssignment = {
...
-helpers.sendFeedbackEmail = (bodyText, assignmentName, email, name) => {
-
+helpers.sendFeedbackEmail = (bodyText, assignmentName, email, name, coachEmail, coachName) => {
   const message = {
     subject: `Feedback on your ${assignmentName}`,
     text: bodyText,
-    from_email: 'support@crisistextline.org',
-    from_name: 'Crisis Text Line',
+    from_email: coachEmail || "support@crisistextline.org",
+    from_name: coachName || "Support",
```
morrismatts: I think in the rest of Elmer we're using single quotes instead of double quotes.

www/canvas/helpers.js
```diff
@@ -236,13 +256,12 @@ helpers.triggersBasedOnAssignment = {
       }),
 };
 
-helpers.sendFeedbackEmail = (bodyText, assignmentName, email, name) => {
-
+helpers.sendFeedbackEmail = (bodyText, assignmentName, email, name, coachEmail, coachName) => {
```
morrismatts: Since we now have so many parameters, you might consider switching this to use ES6 Object Destructuring, which is a fancy way of saying you can pass in the whole object and take out just the parts you like, while also renaming those parts. Like this: `helpers.sendFeedbackEmail(submission)` and then in the definition you can write `helpers.sendFeedbackEmail = ({ comments: bodyText, assignment: assignmentName, email, name, coach: { email: coachEmail, name: coachName }}) => {`

morrismatts: ^ Probably want to double-check my logic there

www/canvas/index.js
```diff
@@ -55,11 +57,18 @@ router.post('/typeformAssignment/:assignment', function (req, res) {
...
       } 
-    // }).then(() => {
-    //   if (helpers.triggersBasedOnAssignment[submission.assignment]) {
-    //     return helpers.triggersBasedOnAssignment[submission.assignment](submission.email);
-    //   }
-    }).then(() => helpers.sendFeedbackEmail(submission.comments, submission.assignment, submission.email, submission.name)
+//     }).then(() => {
+//       if (helpers.triggersBasedOnAssignment[submission.assignment]) {
+//         return helpers.triggersBasedOnAssignment[submission.assignment](submission.email);
+//       }
```
morrismatts: If we don't want this to run can we just delete the commented code? It'll be in git if we need to get it back later...

www/canvas/index.js
```diff
@@ -55,11 +57,18 @@ router.post('/typeformAssignment/:assignment', function (req, res) {
...
+    }).then(() => helpers.sendFeedbackEmail(
+        submission.comments, 
+        submission.assignment, 
+        submission.email, 
+        submission.name,
+        submission.coach.email,
+        submission.coach.name
+      )
```
morrismatts: As per my comment above, we can switch this to just `sendFeedbackEmail(submission)`

canvas.js
mikehow1984: Yeah, we've been using ES6 for a while, so I think we're using an updated version.

Btw, thanks for the heads up. I think I did an ES6 thing in another PR I have (default parameter syntax `function coolFunction(narf, delay = 1000)`). I'll change that

canvas.js
mikehow1984: smh

www/canvas/index.js
mikehow1984: 👍

PR #379, opened 2017-06-21
sends a weekly csv 

mbogor:
#### What's this PR do?
sends a weekly csv to ramneet with people who have logged into home grouped by categories
jobs/okta/oktaLoginsCSV.js
change line 82 to your email and go to town
if you want to run this, contact me for the updated keys file
https://admin.crisistextline.org/jira/browse/AT-724


jobs/okta/oktaLoginsCSV.js
```diff
@@ -0,0 +1,95 @@
+'use strict';
+const nr = require('newrelic');
+const CronJob = require('cron').CronJob;
+const okta = require('../../okta.js');
+const mandrill = require('mandrill-api/mandrill');
+const mandrill_client = new mandrill.Mandrill(KEYS.mandrill.api_key);
+
+// requires an elmer key update
```
apackin: Which key?
I was thinking that we should start updating the keys for elmer before we even merge the PR's that require them... having extra keys can't hurt right?

jobs/okta/oktaLoginsCSV.js
```diff
@@ -0,0 +1,95 @@
...
+	const groupArray = [];
+
+	Object.keys(loginUserObj).forEach(userId=>{
+		groupArray.push(okta.getOktaUserGroups(userId));
+	});
+	return Promise.all(groupArray)
+	.then(groupsArray=>{
+		Object.keys(loginUserObj).forEach((userId, index)=>{
+			let groupsPerUser = groupsArray[index];
+			groupsPerUser.forEach((group, jindex)=>{
```
apackin: I think it might be simpler to do this logic in line 45 via something like:
```const promiseForGroupIds = Object.keys(loginUserObj).map(userId=>{
	return okta.getOktaUserGroups(userId)
              .then(functionThatTurnsGroupPerUserIntoIdsOrElse)
              .then(groupIds => { loginUserObj[userId].group = groupIds }) 
});```

and then `return Promise.all(promiseForGroupIds).then(() => loginUserObj)`

File: jobs/okta/oktaLoginsCSV.js [hunk 3]
```diff
@@ -0,0 +1,95 @@
...
+function newRelicMonit(){
+  nr.createBackgroundTransaction('okta: updatingUsers', getCSV)();
+}
+
+function getCSV(){
+	return oktaLogins()
+	.then(findGroups)
+	.then(usersWithGroups=>{
+		makeACSV(usersWithGroups);
+	})
```
mbogor: makeacsv is not a promise
yep, good catch on the catch

jobs/okta/oktaLoginsCSV.js
```diff
@@ -0,0 +1,95 @@
...
+}
+
+function oktaLogins(){
+	const filter = `action.objectType eq "core.user_auth.login_success" and published gt "2017-06-20T00:00:00.000Z"`;
```
mbogor: yep, this is just for testing, this day is gonna be the date of launch and will be in config

jobs/okta/oktaLoginsCSV.js
```diff
@@ -0,0 +1,95 @@
...
+	return new Promise((resolve, reject) =>{
+		return okta.getOktaLoginEvents(filter)
+		.then(logins=>{
+			const loginUsers = {};
+			logins.forEach(login=>{
+				loginUsers[login.targets[0].id] = {
+					name: login.targets[0].displayName,
+					email: login.targets[0].login,
+					lastLogin: login.published
```
mbogor: the later index will always be the latest login, yes.

jobs/okta/oktaLoginsCSV.js
apackin: .then() accepts functions as the first param - doesn't need to be a promise

jobs/okta/oktaLoginsCSV.js
mbogor: what do you mean by 'more explicitly list'? and yeah, i add keys to prod before i merge the PR's. i left the comment because i wanted to signal that you don't have the keys needed to run this locally

apackin: I meant like saying `// requires key KEYS.okta.groups[group.id] update` instead of `// requires an elmer key update`. nbd either way

PR #383, opened 2017-06-28
added Mini Roleplay to Sherlock ML

mikehow1984:
#### What's this PR do?
Adds the Mini Roleplay assignment in Canvas to Sherlock ML
https://admin.crisistextline.org/jira/browse/AT-710


config.js
```diff
@@ -193,86 +193,161 @@ CONFIG.sherlock = {
...
-      ranking: 2,
-      penalty: -1,
-    },
-    'Didn\'t Introduce Self': {
-      positive: `• Introduced yourself. Fantastic to see you introducting yourself in your first message. Our texters get a few automated messages when they first text us. So it's common for them to wonder if we're human. Sharing your first name goes a long way.`,
-      negative: `• Didn't introduce yourself. We want to give our name in our first message to a texter. Our texters get a few automated messages when they first text us. So it's common for them to wonder if we're human. Sharing your first name goes a long way.`,
-      penalty: -1,
-      ranking: 12,
-      models: [`GCT02`],
+    'Good-Contact Techniques Skill Bite': {
```
mikehow1984: Should we put Sherlock config stuff in its own file? It's getting huge

mikehow1984:
Also [here is the faster throttling](https://github.com/CrisisTextLine/Elmer/pull/383/files#diff-6a1abedd140e08bbbfdf73fabb295ff9R131). The current throttling takes 30 seconds for Google Prediction to predict all the answers in Mini-Roleplay because it uses Bluebird's delay to make one API call, wait for 100ms, then make another (synchronous). This faster throttling simply sends asynchronous API calls every 100ms, regardless of whether the last one has resolved (which makes it run at about 12 seconds).

PR #387, opened 2017-07-05
Update Kaitlyn as head of product

apackin:
#### What's this PR do?
Removes Heather from error handling in Elmer and Adds Kaitlyn. Generalizes name to Head of Product where possible.

https://admin.crisistextline.org/jira/secure/RapidBoard.jspa?rapidView=12&view=detail&selectedIssue=AT-772&quickFilter=44

#### Questions:
Should we set up an email account for Head of Training Product so that we don't have to change this in the code base when this role rotates again? Also then the new Head would have access to all the relevant emails from the previous head.


www/canvas/index.js
```diff
@@ -76,21 +76,21 @@ router.post('/typeformAssignment/:assignment', function (req, res) {
...
+      text: `Hi Head Of Training Product, Due to the error specified in the subject ` + messageText,
       from_email: 'support@crisistextline.org',
       from_name: 'Crisis Text Line',
       to: [{
-          email: 'heather@crisistextline.org',
-          name: 'Heather',
-          type: 'to'
-      }]
+          email: 'kaitlyn@crisistextline.org',
+          name: 'Head Of Training Product',
```
mbogor: didn't we talk about creating a headoftraining ctl email and putting it here instead?

apackin: Yea - He said that he would when I asked about it in stand up, but in the meantime we should prob get these emails going to Kaitlyn instead of Heather?

apackin:
@jaybennett let me know when trainingproduct@crisistextline.org is good to go

PR #388, opened 2017-07-07
excluded feedback for Platform Refresher

mikehow1984:
#### What's this PR do?
Makes it so that the Platform Refresher assignment gives trainees no feedback
https://admin.crisistextline.org/jira/browse/AT-649


www/canvas/index.js
```diff
@@ -61,8 +61,11 @@ router.post('/typeformAssignment/:assignment', function (req, res) {
     //   if (helpers.triggersBasedOnAssignment[submission.assignment]) {
     //     return helpers.triggersBasedOnAssignment[submission.assignment](submission.email);
     //   }
-    }).then(() => helpers.sendFeedbackEmail(submission))
-    .then(() => res.send({message: `Successfully submitted ${submission.assignment} with ${JSON.stringify(submission)}`}))
+    }).then(() => {
+      if (!CONFIG.sherlock.noFeedback.includes(submission.assignment)){
+        helpers.sendFeedbackEmail(submission);
+      }
```
apackin: I feel like this function should return helpers or res.send a different success message if feedback should not be sent. 
Correct me if I'm wrong, but it looks like this way if helpers.sendFeedbackEmail(submission) fails this function still resolves to undefined and continues to send the success message?

mikehow1984: Yeah I think you're right. Changing now

PR #390, opened 2017-07-11
switched to lighter api calls

mbogor:
#### What's this PR do?
lighter okta calls


config.js
```diff
@@ -289,7 +289,7 @@ CONFIG.WiWUserNotes = {
 
 CONFIG.time_interval = {
   // runs every day - WiW Cache
-  cache_refresh: '0 0 0 */1 * *',
+  cache_refresh: '0 0 */2 * * *',
```
mikehow1984: You're changing this to run every 2 hours? (if so, the comment should be changed)

mbogor: fair commentary from an all time favorite michael howard

PR #391, opened 2017-07-18
fixed trainees not getting email from current coach

mikehow1984:
#### What's this PR do?
Trainees who had been assigned to a course in the past were receiving a feedback email from their old coach. This fixes it so that they only get an email from their active coach
https://admin.crisistextline.org/jira/browse/AT-848


mbogor:
👍

PR #393, opened 2017-07-20
if assignment isn't complete, user doesn't get feedback

mikehow1984:
#### What's this PR do?
When a trainee submits an assignment without completing it, they still receive feedback. This makes it so that they do not.

NOTE: It should not be possible to submit an assignment without doing all of the questions. This is an issue that needs to be dealt with in Typeform. It's done by getting a multiple choice question wrong in Typeform, going back, and then scrolling down and hitting "Submit". It's basically the Typeform Konami code. 

https://admin.crisistextline.org/jira/browse/AT-532


kidatsy:
⬆️⬆️⬇️⬇️⬅️➡️⬅️➡️🅱️🅰️

PR #394, opened 2017-07-25
mike/sherlock-wrong-grades/AT-837

mikehow1984:
#### What's this PR do?
Sherlock no longer sends conflicting feedback and new Try It assignments don't send feedback
https://admin.crisistextline.org/jira/browse/AT-837


www/canvas/helpers.js
```diff
@@ -68,13 +68,13 @@ helpers.getFullName = (submission) =>
...
-  new Promise((resolve, reject) => {
+  new Promise((resolve, reject) =>
     canvas.comp.retrieveCoachInfoFromCourseId(submission.courseId)
       .then((coach) => {
         submission.coach = coach;
         resolve(submission);
-      });
-  });
+      })
+);
```
morrismatts: I think this closing paren should be indented once, since it's closing the Promise, right?

www/canvas/helpers.js
```diff
@@ -193,9 +193,14 @@ helpers.formatHTMLresponse = (submission) => {
     submission.submissionHTML += `${breakdownCreator(submission.breakdown)}`;
     submission.comments += comments.creator(commentTally, minWordCount);
   }
-
-  submission.trueGrade = submission.typeformScore + submission.sherlockGradeModifier;
-  if (submission.trueGrade < 0) submission.trueGrade = 0;
+  //pass/fail assignments give a null grade, therefore if an assignment is submitted, it is complete
+  if(submission.typeformScore !== null){
```
morrismatts: space after `if`

www/canvas/helpers.js
```diff
@@ -193,9 +193,14 @@ helpers.formatHTMLresponse = (submission) => {
...
+    submission.trueGrade = submission.typeformScore + submission.sherlockGradeModifier;
+    if (submission.trueGrade < 0) submission.trueGrade = 0;
```
morrismatts: If you're going to add an else clause, you should put brackets around the if clause too, so:
```
if (submission.trueGrade < 0) {
    submission.trueGrade = 0;
} else {
    submission.trueGrade = "complete";
}
```

But actually in this instance I think you can just use a ternary operator, like:
`submission.trueGrade = submission.trueGrade < 0 ? 0 : "complete";`

morrismatts: Also, are we sure we want to be mixing strings and numbers in this `trueGrade` property? It's not always a bad idea, but generally speaking it's better to stick to one type or another, just to reduce confusion down the road/for others.

mikehow1984: The score is interpolated by Canvas into a string when submitted. I leave it a number throughout the code to do calculations. For submissions, it can be either a number grade ("12" or "8.75" or whatever) or a "pass/fail/complete/incomplete". I added this because some assignments are number score, but some are "pass/fail/complete/incomplete", and the weren't being marked as complete because on those assignments, the `submission.typeformScore` was `null`, therefore the `submission.trueGrade` was `null`, so the submitted score has to be "complete/incomplete".

morrismatts: Aha, makes sense to me. Thanks for explaining it!

PR #400, opened 2017-08-07
Manually graded assignments no longer submit a grade

mikehow1984:
#### What's this PR do?
Makes sure manually graded assignments aren't auto-graded in Sherlock


mbogor:
👍

PR #402, opened 2017-08-08
removed auto-grading pass-fail assignments

mikehow1984:
#### What's this PR do?
removes auto-grading for pass-fail assignments (Try It's) on Canvas because coaches still have to manually approve them. There seems to be no way around this in Canvas.


mikehow1984:
Yes

PR #408, opened 2017-08-22
AT-952/added more error checking for error emails

mikehow1984:
#### What's this PR do?
Adds more catch statements and more points where Head of Product is emailed when an error occurs.

Also adds exam questions and answers from the trainee in the error email
https://admin.crisistextline.org/jira/browse/AT-952


canvas.js
```diff
@@ -411,12 +411,14 @@ canvas.activateOrDeactivateEnrollment = function(courseID, enrollmentID, enrollm
 };
 
 canvas.comp = {
-  retrieveUserCourseAssignmentIds: (userEmail, assignment, opt, errFunc) => {
+  retrieveUserCourseAssignmentIds: (userEmail, assignment, opt, errFunc, errFuncAttachment) => {
```
morrismatts: Often a function with this many parameters is a sign that you're trying to do too much in one chunk. Do you think this could be broken down into pieces?

canvas.js
```diff
@@ -427,21 +429,26 @@ canvas.comp = {
...
     })
     .then(assignments => {
-      if (!assignments || assignments.length === 0) throw `Canvas returned 0 assignments`;
-      return assignments[0].id;
+      if (!assignments || assignments.length === 0) throw new Error(`Canvas returned 0 assignments`);
+      else{
+        return assignments[0].id;
+      }
     });
     return Promise.all([promiseUserID, promiseCourseID, promiseAssignmentID]).catch(err => {
```
morrismatts: I'm kind of confused whats going on here. It looks like this `Promise.all` is to await all of these promises, but each of the promises already relies on the previous promise in the code above (e.g,. `const promiseCourseID = promiseUserID.then(...)`. Is there some reason we've set it up this way? Wouldn't it be a lot cleaner to just chain the promises together and have the `catch` at the end?

morrismatts: Btw, I see that you didn't write the code in this commit, you're just updating it, but I'm asking anyway! It might be that cleaning this up is a job for a bigger story.

www/canvas/helpers.js
```diff
@@ -55,75 +55,89 @@ helpers.constructAssignmentObj = (req, assignment) => {
...
     .then((data) => {
-      submission.name = data.length ? data[0].name : submission.email;
-      resolve(submission);
-    })
+      if (!data.length){
+        throw new Error();
+      }
+      submission.name = data[0].name;
+      resolve([submission, errFunc]);
+    }).catch(err => errFunc(`Error finding info for user ${submission.email}\n`, submission, err))
```
morrismatts: I think `.catch(...` should be on a new line, like `.then(...` above.

www/canvas/helpers.js
```diff
@@ -55,75 +55,89 @@ helpers.constructAssignmentObj = (req, assignment) => {
...
-helpers.getCoachInfo = (submission) => 
+helpers.getCoachInfo = ([submission, errFunc]) => 
   new Promise((resolve, reject) =>
     canvas.comp.retrieveCoachInfoFromCourseId(submission.courseId)
       .then((coach) => {
         submission.coach = coach;
-        resolve(submission);
-      })
+        resolve([submission, errFunc]);
+      }).catch(err => errFunc(`Error finding coach info for user ${submission.email}\n`, submission, err))
```
morrismatts: `.catch` on new line

www/canvas/helpers.js
```diff
@@ -296,4 +309,12 @@ helpers.sendFeedbackEmail = (submission) => {
   });
 };
 
+helpers.errorQandA = (submission) => {
+  return submission.sortedQuestions.map((q, idx) => {
+    const qWithBolding = q.question.replace('Texter:', '<b>Texter:</b>');
+    let html = `<h2>${idx}: ${qWithBolding}</h2><h4><b>Crisis Counselor: </b><i>${q.answer}</i></h4>`;
+    return html;
```
morrismatts: If you're not modifying the variable it should be a `const`, but you could also just return it directly.

canvas.js
```diff
@@ -411,12 +411,14 @@ canvas.activateOrDeactivateEnrollment = function(courseID, enrollmentID, enrollm
...
 canvas.comp = {
-  retrieveUserCourseAssignmentIds: (userEmail, assignment, opt, errFunc) => {
+  retrieveUserCourseAssignmentIds: (userEmail, assignment, opt, errFunc, errFuncAttachment) => {
     // request canvas user with this email address and find their userId
     const promiseUserID = canvas.getUsers(userEmail, opt.delay)
     .then((users) => {
-      if (!users || users.length === 0) throw 'No user was found in Canvas for that email.';
-      return users[0].id;
+      if (!users || users.length === 0) throw new Error('No user was found in Canvas for that email.');
+      else{
```
morrismatts: space between `else` and `{` (same thing below)

canvas.js
mikehow1984: I think it was written this way to be asynchronous because the calls don't depend on each other to resolve, so it's quicker to let them all fly and `Promise.all` them.

canvas.js
mikehow1984: I mean, the first 3 parameters are necessary. The function itself is a combo between two Canvas API calls. The last two parameters are for the error email. Not sure if this can be broken down further without defeating the purpose.

mikehow1984: the `errFuncAttachment` is optional, but I'm also going to make `errFunc` optional too

morrismatts: 👍 As long as you feel like it's necessary, I'm cool with that. Parameter count is just a good rule of thumb, not a law!

PR #409, opened 2017-08-25
Assaf/parameterize for international/at 943

apackin:
#### What's this PR do?
Moves hardcoded domains and contact info to config
Fixes a bunch of linter errors (where I couldn't help myself)

https://admin.crisistextline.org/jira/browse/AT-943

#### Questions:
There's a lot of code in that is already out of use/probably doesn't work anymore. I deleted some of it but there's still:
Everything to do with ctl online
The gsheets integration in the data route
Should I go ahead and remove those too?

I also didn't spend much time looking into WiW related files because they should be deprecated by McFly - anything in /Scheduling that you think could make it to international?

I move `absenteeUsers: 'https://app.crisistextline.org/ajax/stats/absentees',` into KEYS.platform.urls because that's were we are keeping other platform api routes. I'm not sure if that actually has to be 'secret'. Also wanted to highlight that because of that *THIS PR INCLUDES A KEY.js CHANGE*


apackin:
This is ready for review @mikehow1984 @mbogor . Thank you!

PR #411, opened 2017-08-29
AT-1047/fix-sherlock-error-handling

mikehow1984:
#### What's this PR do?
Routes constantly retry without a 2XX status sent. This PR forces a 202 status when a trainee submits a Sherlock assignment that generates an error.
https://admin.crisistextline.org/jira/browse/AT-1027


www/app.js
```diff
@@ -3,6 +3,7 @@
 //REQUIRING KEYS BEFORE CONFIG BECAUSE CONFIG DEPENDS ON KEYS
 global.KEYS = require('../keys.js');
 global.CONFIG = require('../config');
+global.USERS_WITH_SHERLOCK_ERRORS = {};
```
apackin: You made it global, but are also importing it into the files you're using it in?
I think the latter makes more sense in this case.

mikehow1984: Yes, actually. I think I made that change before I decided to make the sherlockErrorHandler module

apackin:
You actually may need to throw the error to get it to not continue with the rest of the chain if that's the goal

apackin:
It actually looks like almost all of those helpers are calling promises in the from `canvas.js`. Do we need to wrap them in `new Promise((resolve, reject) => {` ?

www/canvas/helpers.js
```diff
@@ -60,86 +61,92 @@ helpers.constructAssignmentObj = (req, assignment) => {
...
-    const msg = `There was a problem retrieving the courseId for user ${submission.email}\n${err.stack}\n`;
-    CONSOLE_WITH_TIME(msg);
-    errFunc(msg, submission, err);
+    const message = `There was a problem retrieving the courseId for user ${submission.email}\n${err.stack}\n`;
+    CONSOLE_WITH_TIME(message);
+    sherlockErrorHandler.errorCodeSender(submission, res, message, err, emailErrFunc);
   });
 
-helpers.getAssignmentType = ([submission, errFunc]) => new Promise((resolve, reject) => {
+helpers.getAssignmentType = ([submission, res, emailErrFunc]) => new Promise((resolve, reject) => {
```
apackin: To explain my higher level comment, I think if you make this function:
`helpers.getAssignmentType = (submission) =>
  canvas.getAssignments(submission.courseId, submission.assignment)
  .then((assignment) => {
    submission.assignmentType = assignment[0].grading_type;
    return (submission);
  })
  .catch(err => {
    const msg = `There was a problem retrieving the assignment type for user ${submission.email} in course ${submission.courseId}\n`;
    throw (msg);
  });
`

Then the thrown msg will be caught in the higher level promise chain in www/canvas/index.js so you only have to call sherlockErrorHandler.errorCodeSender there.

mikehow1984: Yeah, I see. I could also throw the reject() instead of the throw. This completely gets rid of the need to even split out SherlockErrorHandler 👍

apackin: Yup :-)
I opted away from the reject because I don't think they need to be wrapped in `new Promise((resolve, reject) => {` at all

www/canvas/helpers.js
```diff
@@ -60,86 +60,85 @@ helpers.constructAssignmentObj = (req, assignment) => {
...
-        throw new Error();
-      }
-      submission.name = data[0].name;
-      resolve([submission, errFunc]);
-    })
-    .catch(err => errFunc(`Error finding info for user ${submission.email}\n`, submission, err))
+helpers.getFullName = (submission) => 
+  canvas.getUsers(submission.email)
+  .then((data) => {
+    submission.name = data[0].name;
```
apackin: Should `if (!data.length){` still be an error?

www/canvas/helpers.js
```diff
@@ -60,86 +60,85 @@ helpers.constructAssignmentObj = (req, assignment) => {
...
-      });
-  }).catch((err) => {
-    const msg = `There was a problem retrieving the courseId for user ${submission.email}\n${err.stack}\n`;
-    CONSOLE_WITH_TIME(msg);
-    errFunc(msg, submission, err);
-  });
+helpers.getCourseId = (submission) => 
+  canvas.comp.retrieveCourseEnrollmentIdsFromEmail(submission.email)
+    .then((courseIds) => {
+      if (!courseIds.active.length) reject(`Tried to retrieve active course info for ${submission.email} but user is not enrolled in any active courses!`);
```
apackin: Reject -> throw

www/canvas/helpers.js
```diff
@@ -60,86 +60,86 @@ helpers.constructAssignmentObj = (req, assignment) => {
...
+helpers.getCourseId = (submission) => 
+  canvas.comp.retrieveCourseEnrollmentIdsFromEmail(submission.email)
+    .then((courseIds) => {
+      if (!courseIds.active.length) throw `Tried to retrieve active course info for ${submission.email} but user is not enrolled in any active courses!`;
+      submission.courseId = courseIds.active[0];
+      return submission;
+    })
+    .catch((err) => {
+      const message = `There was a problem retrieving the courseId for user ${submission.email}\n${err.stack}\n`;
+      CONSOLE_WITH_TIME(message);
```
apackin: We're already doing this CONSOLE_WITH_TIME in the top level catch

www/canvas/index.js
```diff
@@ -80,11 +80,14 @@ router.post('/typeformAssignment/:assignment', timeout('60s'), function (req, re
       if (!usersThatHaveErrors[submission.email]) {
         usersThatHaveErrors[submission.email] = 1;
         res.status(500).send({message});
+        logErrEmailHeadTrainingProduct(message, submission, err);
       } else if (usersThatHaveErrors[submission.email] === CONFIG.sherlock.retrySubmission) {
+        delete usersThatHaveErrors[submission.email];
```
apackin: I'm not sure that we want to delete these: Once we email that a user is having an issue do we want to send emails again for each different issue that user has?

mikehow1984:
Kaitlyn said that she wants the email to be sent again if they keep trying to submit the assignment.

PR #420, opened 2017-10-06
Mariya/mcfly/at 1050

mbogor:
we need to update keys on prod before me merge this ish


jobs/scheduling/mcflyFirstShift.js
```diff
@@ -0,0 +1,65 @@
...
+	const params = {
+		uri: KEYS.mcfly.urls.getShifts,
+		headers: {'Authentication': KEYS.mcfly.apiKey},
+		method: 'GET',
+		qs: {createdAt: moment().utc().subtract(1, 'day').format('YYYY-MM-DD HH:MM:SS'), include_users: true},
+		json: true,
+	};
+	return request(params)
+	.then(response => {
+		if (!response || response.length) return;
```
apackin: Should this be `!response || !response.length` ?

jobs/scheduling/mcflyFirstShift.js
```diff
@@ -0,0 +1,65 @@
...
+		const creditTime = [];
+		response.forEach(shift => {
+			const user = shift.user;
+			if (user.role=='trainee') creditTime.push(user.loginEmail);
```
apackin: ===

apackin: I'm not sure how many shifts each user might have, but it we're expecting more than 1 to be common it could make sense to make creditTime an object so we don't add the same one multiple times.

jobs/scheduling/mcflyFirstShift.js
```diff
@@ -0,0 +1,65 @@
...
+
+new CronJob(CONFIG.time_interval.cron_twice_per_day, newRelicMonit, null, true);
+
+function newRelicMonit(){
+	nr.createBackgroundTransaction('scheduling: mcflyFirstShift', main)();
+}
+
+newRelicMonit();
+
+function main(){
```
apackin: Can we use es6 `const main = () => ` for these function declarations instead?

jobs/scheduling/mcflyFirstShift.js
```diff
@@ -0,0 +1,65 @@
...
+			if(!responses || responses.length) return;
+		// responses is [userID, courseID, assignmentID] so we spread it
+			return canvas.updateUserGrade(...responses, 'pass');
+		})
+		.catch(function(err) {
+		CONSOLE_WITH_TIME(`MCFLY JOB: Finding the user with the email ${email} in Canvas failed: ${err}`);
+		});
+	}
+
+	const promises = emailArray.map(email=>{return canvasTime(email)});
```
apackin: `email=>{return canvasTime(email)}` -> `email => canvasTime(email)`

jobs/scheduling/mcflyFirstShift.js
```diff
@@ -0,0 +1,65 @@
...
+
+
+function canvasCredit(emailArray){
+	if (!emailArray.length) return;
+
+	function canvasTime(email){
+		return canvas.comp.retrieveUserCourseAssignmentIds(email, CONFIG.canvas.assignments.scheduledShifts, {}, CONSOLE_WITH_TIME)
+		.then(responses => {
+			if(!responses || responses.length) return;
+		// responses is [userID, courseID, assignmentID] so we spread it
```
apackin: Love the note 👍

jobs/scheduling/mcflyFirstShift.js
```diff
@@ -0,0 +1,65 @@
...
+}
+
+
+function canvasCredit(emailArray){
+	if (!emailArray.length) return;
+
+	function canvasTime(email){
+		return canvas.comp.retrieveUserCourseAssignmentIds(email, CONFIG.canvas.assignments.scheduledShifts, {}, CONSOLE_WITH_TIME)
+		.then(responses => {
+			if(!responses || responses.length) return;
```
apackin: Also think this should prob be `!response || !responses.length`

